PR #1, opened 2022-04-14
Update submodule url to ferdium-recipes 

nathanaelhoun:
#### Checklist
- [x] Use `ferdium-recipes` submodule
- [x] Added a step to install `node-gyp` in windows ci
- [x] CI badge in README.md not updated (done in #3)

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -310,7 +310,7 @@ jobs:
         with:
           version: 6.24.1
       - name: Install node-gyp
-        run: npm install -g node-gyp
+        run: npm install -g node-gyp@8.4.1
       - name: Install node dependencies
         uses: nick-invision/retry@v2.4.0
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/builds.yml
```diff
@@ -309,6 +309,8 @@ jobs:
         uses: pnpm/action-setup@v2.0.1
         with:
           version: 6.24.1
+      - name: Install node-gyp
+        run: npm install -g node-gyp
```
vraravam, line 313: I would strongly prefer that we specify a version. Looking at [this line](https://github.com/ferdium/ferdium-app/blob/develop/package.json#L186) - we should go with `8.4.1`

vraravam (changes requested):

.github/workflows/builds.yml
nathanaelhoun: Oops of course, forgot :facepalm: 
```suggestion
        run: npm install -g node-gyp@8.4.1
```

vraravam (approved):

vraravam:
LGTM!

Issue #17, opened 2022-04-20
Whatsapp and Discord (maybe others) logout after some time of no usage in Ferdium


vraravam:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Steps to reproduce

1. Add Whatsapp as a service into Ferdium
2. Login after scanning the QR code
3. Do not use Whatsapp in Ferdium for some time (15+ mins?)
4. You are logged out

### Debug link

na

### Additional information

This seems to have been an after-effect of the upgrade from electron 15.3 to 16 and beyond. Might need to investigate if we have to do some special processing. We are already using the persisted sessions for all webviews, but this doesn't seem to work atm

vraravam:
Might be related to this? https://github.com/electron/electron/issues/959#issuecomment-985964273

SpecialAro:
I'm having a similar problem with Whatsapp whenever I close and open Ferdium. It seems that credentials are not persistent.
Tho, for messenger everything is fine.

kris7t:
My guess would be something related to https://github.com/electron/electron/issues/31689

We should check whether there's anything in the webview preload script (https://github.com/ferdium/ferdium-app/blob/develop/src/webview/recipe.js and all the stuff it imports) that tries to access `localStorage`.

The most likely problem is the `require('debug')` call (in `recipe.js` and probably in numerous scripts it imports), because the `debug` NPM package attempts to use `localStorage` to save the current log level: https://www.npmjs.com/package/debug#user-content-browser-support Due to the electron bug above, user data in the service will be lost.

mmajewsk:
I have the same issue in 6.7.0 and solemly occuring in whatsapp and discord

uroybd:
This issue still persists.

odie16:
Happening to me too with WhatsApp on version 6.7.2.

theoratkin:
I recently (in the last 2-3 weeks) started having this problem with Discord, it logs me out on every restart. Is this a problem with Ferdium or Electron?

PR #23, opened 2022-04-22
Turn off usage of 'debug' npm package using with electron-16+

vraravam:
#### Pre-flight Checklist

#### Description of Change

As discussed in #17, temporarily turning off debugging using the 'debug' npm module.
Please note that I have not trimmed this list of files - they all seem logically connected - and when I turned off and restart Ferdium - my discord was still working!

#### Checklist

- [x] `npm test` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "csstype": "^3.0.9",
         "darkreader": "^4.9.39",
         "dbus-next": "^0.10.2",
-        "debug": "^4.3.2",
+        "debug": "4.3.4",
         "electron-dl": "^3.2.1",
         "electron-find": "^1.0.7",
         "electron-react-titlebar": "^1.1.0",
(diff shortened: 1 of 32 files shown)


kris7t:
This is way too extreme, I think. We don't have to remove all uses of `debug`, only those that run in the `preload` script of our `webview`. Of course, figuring out which of the usages are they is a bit difficult, because main/renderer/preload scripts aren't kept separate in the ferdium codebase, unfortunately.

vraravam:
yes, unfortunately, there's no other way I could think of. As soon as file `A` imports file `B`, if the `require('debug')` is evaluated in file `B`, then we have the same problem, isn't it?

kris7t:
We could try to create a shim which we import everywhere instead of `debug`, that check whether we're in the preload environment and only requires `debug` if we aren't (it can just return the `console.log` method in the preload, for example). This we at least have debugging support in the rest of the app.

Something like

```javascript
export default function debug(namespace) {
  if (globalThis.ferdiumPreload) {
    return console.log;
  }
  return require('debug')(namespace);
}
```

saved as `preloadSafeDebug.js`, and then replace each use of `require('debug')` with `require('../preloadSafeDebug')` or something like that. Then `webview.js` could set `globalThis.ferdiumPreload = true;` to avoid pulling in `debug`.

But this is a pretty rough idea, and I'll have to try it to see whether it works.

vraravam:
I thought about this - but, im not enough of a js dev to have implemented this.

kris7t:
I'll try to do the alternative fix on top of a _revert_ of this. Thus, we can push out a working nightly now, and get back debugging support later

kris7t (approved):

mydarkstar:
@kris7t I like the shim idea!
It could also use some currying to display the namespace for each `console.log` call, along the lines of:
```js
return (...params) => console.log(`[${namespace}]`, ...params);
```

Also, just as a side note. I'm not entirely sure how Node implements this, but could there be a performance issue with calling `require` on every function call?

kris7t:
We'd only call `require` on each `debug` function call, but that's usually only called once per module (to initialize a new logger for the namespace), so I wouldn't be too afraid of it.

Adding the prefix is a nice idea!

dqos:
This looks good to me, can we proceed with merging? Then we can roll out nightly.8 asap.

PR #24, opened 2022-04-22
Updates the build script for Windows (Powershell)

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Updates the build script for Windows (Powershell)

#### Motivation and Context
The build script for Windows (Powershell) needed update in order to be coherent with the existing shell script (https://github.com/ferdium/ferdium-app/blob/develop/scripts/build-unix.sh).

#### Checklist

#### Release Notes
none


scripts/build-windows.ps1
```diff
@@ -18,28 +14,97 @@ $env:CSC_IDENTITY_AUTO_DISCOVERY = $false
...
+   expected: [v$EXPECTED_NODE_VERSION]
+   actual  : [$ACTUAL_NODE_VERSION]"
+}
+
+# Checking npm version
+$EXPECTED_NPM_VERSION = (Get-Content package.json | ConvertFrom-Json).engines.npm 
+$ACTUAL_NPM_VERSION = (npm -v)
+
+if ( $EXPECTED_NPM_VERSION -ne $ACTUAL_NPM_VERSION) {
+  Write-Host "You are not running the expected version of npm!
```
vraravam: Shouldn't this also refer to `fail_with_docs`?

vraravam (changes requested):

scripts/build-windows.ps1
SpecialAro: In this case it works just as a warning. This is because we previously test if it is installed (in line 45) and we use the correct version in line 123 (npm i -gf npm@8.7.0).

But if it is more coherent, we can exit the script there as well.


Just noticed that lines 123 and 124 should change as well... They need to be referenced to package.json. I'll commit those changes before you close this PR

scripts/build-windows.ps1
```diff
@@ -18,28 +14,97 @@ $env:CSC_IDENTITY_AUTO_DISCOVERY = $false
...
+Test-CommandExists node "Node is not installed"
+Test-CommandExists npm "npm is not installed"
+Test-CommandExists python "Python is not installed"
+# NEEDS proper way to CHECK MSVS Tools
+
+# Checking node.js version
+$EXPECTED_NODE_VERSION = (cat .nvmrc)
+$ACTUAL_NODE_VERSION = (node -v)
+
+if ( "v$EXPECTED_NODE_VERSION" -ne $ACTUAL_NODE_VERSION) {
```
vraravam: nitpick: spaces around parens - please be consistent

scripts/build-windows.ps1
```diff
@@ -18,28 +14,97 @@ $env:CSC_IDENTITY_AUTO_DISCOVERY = $false
...
+
+if ( $EXPECTED_NPM_VERSION -ne $ACTUAL_NPM_VERSION) {
+  Write-Host "You are not running the expected version of npm!
+   expected: [$EXPECTED_NPM_VERSION]
+   actual  : [$ACTUAL_NPM_VERSION]"
+  Write-Host "Changing version of to [$EXPECTED_NPM_VERSION]"
+}
+
+# Checking python version
+$EXPECTED_PYTHON_VERSION = "3.6"
```
vraravam: python version being discussed earlier in the thread was `3.10.4`

vraravam (changes requested):

scripts/build-windows.ps1
SpecialAro: Not sure if is in fact a dependency. I looked for the image I posted on discord and python 3.6 was the lowest possible on the error I got. Shall I keep 3.6 or force at least 3.7?

SpecialAro: unknown

scripts/build-windows.ps1
```diff
@@ -55,7 +120,7 @@ if ( $env:CLEAN -eq "true" )
 }
 
 # Ensure that the system dependencies are at the correct version
-npm i -gf npm@8.7.0
+npm i -gf npm@$EXPECTED_NPM_VERSION
 npm i -gf pnpm@6.32.8
```
nathanaelhoun: With the same code you added above (I don't know powershell), we can get the PNPM version aswell
```suggestion
npm i -gf pnpm@$((Get-Content recipes\package.json | ConvertFrom-Json).engines.pnpm)
```

scripts/build-windows.ps1
vraravam: i think the `CONTRIBUTING.md` has `3.10.4` - which works on my machine (macos)

nathanaelhoun: Version referenced in `CONTRIBUTING.md` is python 3.10.4, so we should update here or in `CONTRIBUTING.md` with a version that works :-)
EDIT: too slow :smile:

SpecialAro: Yeah, it was me who added that on CONTRIBUTING.md . I can try to build everything with python 3.6 and I will tell you if it worked or not, so as to find that if it is a requirement or not.

What do you think about this?

scripts/build-windows.ps1
SpecialAro: Yes we can, but the package pnpm is not a dependency available on package.json (in the discord we were discussing the possibility of having it or not has such). There is currently an PR open on that I think, to switch from npm to pnpm on the entire project but is a work in progess (https://github.com/ferdium/ferdium-app/pull/26).

nathanaelhoun: I should have underlined that: this line is reading from [**recipes**/package.json](https://github.com/ferdium/ferdium-recipes/blob/master/package.json#L7), which does use `pnpm` (and that is why we use `pnpm` in the script)

scripts/build-windows.ps1
vraravam: yes, we can go with whatever works across all 3 OSes that we usually build for.

If its the lower version of python, im fine to downgrade - not a problem. Just that we need to be consistent across all 3 OSes so that we can uncover similar issues

scripts/build-windows.ps1
vraravam: that sounds correct @nathanaelhoun

vraravam: incorporated in the latest commit. tx @nathanaelhoun

scripts/build-windows.ps1
vraravam: in the latest commit (for now), I have set the python version to `3.10.4`. please confirm if we want to go with this or the lower one. tx

SpecialAro: The expression that I have on the code implies 3.6 or higher, which means that it would work with whatever version that is higher, so it wouldn't break if we are using the latest python.

scripts/build-windows.ps1
SpecialAro: Wow! That way is fantastic awesome!

scripts/build-windows.ps1
vraravam: yes, but as i had mentioned in other threads, im not in favor of keeping it open-ended, there will be issues in the future where things break randomly on different machines due to versions being different. This has happened in the past, and I prefer to remain locked. The upgrade needs to be an explicit intentional action from a dev and then it can be monitored for all OSes.

vraravam: Resolving for now - we can touch back on this specific topic in a later PR/thread

vraravam (approved):

vraravam:
@all-contributors please add @SpecialAro for infrastructure

PR #29, opened 2022-04-22
Preload safe debug shim

kris7t:
#### Pre-flight Checklist

#### Description of Change
This patch aims to restore some debug functionality by introducing a shim. The shim detect whether if it is being introduced in a preload script where the electron but would be triggered, and falls back to a simple replacement for debug. However, in the main and renderer processes, where a preload script is not being used, we still get full debug functionality.

In this way, a module can be used both in a preload script and outside of it, while still preserving debug functionality whenever possible.

#### Motivation and Context
In #23 we removed usages of the debug package due to an electron bug.
However, this hard-coded `console.log` instead of a more comfortable logger in some places.

#### Checklist

- [x] `npm test` passes

#### Release Notes

none

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "csstype": "^3.0.9",
         "darkreader": "^4.9.39",
         "dbus-next": "^0.10.2",
-        "debug": "4.3.4",
+        "debug": "^4.3.2",
         "electron-dl": "^3.2.1",
         "electron-find": "^1.0.7",
         "electron-react-titlebar": "^1.1.0",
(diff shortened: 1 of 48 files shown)


package-lock.json
```diff
@@ -33,7 +33,7 @@
         "csstype": "^3.0.9",
         "darkreader": "^4.9.39",
         "dbus-next": "^0.10.2",
-        "debug": "4.3.4",
+        "debug": "^4.3.2",
```
vraravam: can we please use `4.3.3` as the specific version if that's what you want to use? I don't prefer open-ended version definitions - since that can then cause weird errors across machines.

package.json
```diff
@@ -68,7 +68,7 @@
     "csstype": "^3.0.9",
     "darkreader": "^4.9.39",
     "dbus-next": "^0.10.2",
-    "debug": "4.3.4",
+    "debug": "^4.3.2",
```
vraravam: can we please use `4.3.3` as the specific version if that's what you want to use? I don't prefer open-ended version definitions - since that can then cause weird errors across machines.

vraravam: Also, is there a reason why we are downgrading from 4.3.4 to 4.3.3?

vraravam (changes requested):

package.json
kris7t: Ah, no, the downgrade just came from reverting your commit. I'll put the bump back.

vraravam (approved):

PR #103, opened 2022-05-05
Do not allow http urls that end in '/' (fixes #102)

vraravam:
#### Pre-flight Checklist

1. Please remember that if you are logging a bug for some service that has _stopped working_ or is _working incorrectly_, please log the bug [here](https://github.com/ferdium/ferdium-recipes/issues)
2. If you are requesting support for a **new service** in Ferdium, please log it [here](https://github.com/ferdium/ferdium-recipes/pulls)
3. Please remember to read the [self-help documentation](https://github.com/ferdium/ferdium-app#troubleshooting-recipes-self-help) - in case it helps you unblock yourself for issues related to older versions of recipes that were installed on your machine. (These will get automatically upgraded when you upgrade to the newer versions of Ferdium, but to get new recipes between Ferdium releases, this documentation is quite useful.)
4. Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

#### Description of Change

Add validation when checking for URLs to also validate that it doesn't end in `/`

#### Motivation and Context

Fixes #102 

#### Screenshots

Screenshot 2022-05-05 at 6 22 04 PM
Screenshot 2022-05-05 at 6 22 17 PM

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (`npm run prepare-code`)
- [x] `npm test` passes
- [x] I tested/previewed my changes locally

Issue #102, opened 2022-05-05
Bug: If custom server URL ends in "/", app fails to login and signup


amatriain:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.0.0.nightly.26

### What Operating System are you using?

Other Linux

### Operating System Version

arch linux (rolling release, latest updates as of 5/6/2022)

### What arch are you using?

x64

### Expected Behavior

When using a custom server, login and signup from the app should work even if the server URL is terminated with a "/" character.

### Actual Behavior

When the custom server URL is terminated by a "/" (e.g. "https://ferdium.some.site/") neither login nor signup work from the desktop app. No error is displayed, but clicking on the "Sign in" or "Create account" buttons does nothing.

Removing the trailing "/" from the URL (e.g. "https://ferdium.some.site") makes login and signup work again.

### Steps to reproduce

1. Open ferdium app
2. Go to "Change server"
3. In "Server" dropdown, choose "Custom"
4. In "Custom server" enter the URL of a working ferdium-server instance with a trailing /
5. Click "Submit" button
6. Click "Login to your account"
7. Enter credentials and click "Sign in" button
8. Nothing happens
9. Repeat the steps above to remove the trailing / from the server URL and try to log in
10. Login now works

### Additional information

My guess is the app uses string concatenation to create the login URL, and when the custom server URL has a trailing / the resulting login URL has a double // somewhere.

amatriain:
Replied in the PR: https://github.com/ferdium/ferdium-app/pull/103#issuecomment-1119149633

PR #105, opened 2022-05-06
New fix for fixing URLs with '//'

vraravam:
#### Pre-flight Checklist

#### Description of Change

Reverted previous fix from #103 and added new changes as per [comment](https://github.com/ferdium/ferdium-app/pull/103#issuecomment-1119149633)

#### Motivation and Context

Should now be a better fix for #102

#### Checklist

- [x] `npm test` passes
- [ ] I tested/previewed my changes locally

### DIFF
--- a/src/helpers/url-helpers.ts
+++ b/src/helpers/url-helpers.ts
@@ -15,15 +15,16 @@ export function isValidExternalURL(url: string | URL) {
   } catch {
     return false;
   }
-  if (url.toString().endsWith('/')) {
-    return false;
-  }
 
   const isAllowed = ALLOWED_PROTOCOLS.includes(parsedUrl.protocol);
 
   debug('protocol check is', isAllowed, 'for:', url);
 
   return isAllowed;
+}
+
+export function fixUrl(url: string | URL) {
+  return url.toString().replaceAll('//', '/').replace('http:/', 'http://').replace('https:/', 'https://').replace('file:/', 'file://');
 }
 
 export function isValidFileUrl(path: string) {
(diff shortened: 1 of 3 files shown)


vraravam:
Unfortunately, the changes seem to break the app while bootup itself - I dont use Ferdium with any account, but the initial login screen itself doesn't come up for me. I'm unable to take this forward (have travel plans this weekend). If someone can continue and find the fix, i would really appreciate it

nathanaelhoun:
Instead of fixing the URL everywhere we use it, I propose two alternatives approaches:
1. Modify the form submission to remove a trailing slash ([here](https://github.com/ferdium/ferdium-app/blob/develop/src/components/auth/ChangeServer.js#L124)) with a 
   ```js
   this.form.$('customServer').value = this.form.$('customServer').value.replace(/\/$/, "")
   ```
2. Modify the [`apiBase()` and `termsBase()` methods](https://github.com/ferdium/ferdium-app/blob/develop/src/api/apiBase.ts#L14) to remove the trailing slash before returning the value

vraravam:
> 2\. Modify the [`apiBase()` and `termsBase()` methods](https://github.com/ferdium/ferdium-app/blob/develop/src/api/apiBase.ts#L14) to remove the trailing slash before returning the value

OMG! Thanks for catching the rabbit hole that I was digging!
Will fix both in the next commit in a few mins

vraravam:
@nathanaelhoun - if you can review this and approve, we can merge for the next nightly

src/helpers/url-helpers.ts
```diff
@@ -26,6 +23,11 @@ export function isValidExternalURL(url: string | URL) {
   return isAllowed;
 }
 
+export function fixUrl(url: string | URL) {
+  return url.toString().replaceAll('//', '/').replace('http:/', 'http://').replace('https:/', 'https://').replace('file:/', 'file://');
```
nathanaelhoun, line 27: I think this isn't needed now as we store the url correctly :-)

nathanaelhoun (approved):
Looks good to me!

src/helpers/url-helpers.ts
vraravam: yes - but, i felt its a safety net in case of new code where there's `//`. Do you still want me to remove it? I would prefer to keep it for now - unless it causes trouble in the future

Issue #157, opened 2022-05-16
Bug: Services are loading infinitely in some conditions


MagicLegend:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.0.0-nightly.37

### What Operating System are you using?

Ubuntu

### Operating System Version

Linux <name> 5.13.0-41-generic #46~20.04.1-Ubuntu SMP Wed Apr 20 13:16:21 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### What arch are you using?

x64

### Expected Behavior

It works.

### Actual Behavior

6 out of 8 newly added services get stuck at `Loading <service>`.

### Steps to reproduce

1. Install Ferdium
2. Add multiple services
3. Some should get stuck on loading

### Screenshots

![image](https://user-images.githubusercontent.com/3169104/168537000-ca262412-4a53-4cd1-96a2-e5883f1a03a5.png)
![image](https://user-images.githubusercontent.com/3169104/168537015-f6e6d5a0-2c81-4c94-813e-d687c297f8fc.png)
![image](https://user-images.githubusercontent.com/3169104/168537031-ce6fe102-ac37-4543-ab7a-e7af311898f0.png)

### Additional information

While writing this issue it seems that Gmail and Outlook got un-stuck.
Reloading the service was not enough (although the service did refresh), only reloading Ferdium fixed it.

SpecialAro:
I was able to replicate this issue when adding 7 services. Only 1 got stuck (Messenger) and the reload of the service didn't work - only the reloading of Ferdium (as @MagicLegend indicates) solved the problem.

Evan-aja:
> I was able to replicate this issue when adding 7 services. Only 1 got stuck (Messenger) and the reload of the service didn't work - only the reloading of Ferdium (as @MagicLegend indicates) solved the problem.

I have encountered this issue as well a few days ago. in my setup there is __one enabled__ (WhatsApp) service and __six disabled__ (2 Gmails, 2 Elements, Discord, and Telegram) at all time. 

And the loading indicator bug shows up randomly at startup and i don't know why. My way of circumventing this issue is by holding any nightly update until the error occurs, updates when it occurs, hold it again from updating, rinse and repeat (tested on both __ArchLinux__ and __OpenSUSE Tumbleweed__).

vraravam:
[...]

Issue #239, opened 2022-06-12
Feature: Progress Indicator for page loading and downloads


pepper3k:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

When I refresh the page, or load a new page within a service, sometimes because the webapp is slow I'm not sure if anything is happening.

The same problem occurs when I download a file within a service.

### Proposed Solution

I solution to this would be to create some sort of progress indicator for page loading and downloads.

Example for replacing the service logo while loading:

https://i0.wp.com/css-tricks.com/wp-content/uploads/2021/08/s_2A9C470D38F43091CCD122E63014ED4503CAA7508FAF0C6806AE473C2B94B83E_1627522653545_loadinfo.gif?resize=200%2C200&ssl=1

A couple of examples for a download progress bar:

https://codepen.io/shankarcabus/pen/kOEYBO
https://codepen.io/aorcsik/pen/JXMvar

The information displayed would be path/filename, downloadKB/totalKB, download speed, download link, etc..
Also, there would be clickable options to open file, or open file location.
Al

### Alternatives Considered

Nothing

vraravam:
Could I request you to please separate out the 2 items into 2 feature requests?

The first one: there's already a loader in the app, and I think it (or a similar loader to what you have linked) can be implemented in a simple manner.

The download manager ui that you describe is much more complex, and will take time.

HenrijsS:
@SpecialAro What do you think about a more subtle progress bar at the top?

Here's a pretty simple library for it: https://kimmobrunfeldt.github.io/progressbar.js/

For example:
Animate the bar to 0.4 when starting navigation and to 1.0 when finished and hide it. (Like some SPA Frameworks do it).

SpecialAro:
> @SpecialAro What do you think about a more subtle progress bar at the top?
> 
> Here's a pretty simple library for it: https://kimmobrunfeldt.github.io/progressbar.js/
> 
[...]

PR #248, opened 2022-06-14
Feature: Progress Indicator for page loading

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Feature: Progress Indicator for page loading

#### Motivation and Context
Feature: Progress Indicator for page loading. Helps fix partially #239.

#### Screenshots
Animation

#### Checklist

- [x] `npm test` passes

#### Release Notes
Feature: Progress Indicator for page loading

### DIFF
--- a/src/components/services/content/ServiceView.js
+++ b/src/components/services/content/ServiceView.js
@@ -113,6 +113,9 @@ class ServiceView extends Component {
               !service.isServiceAccessRestricted && (
                 <WebviewLoader loaded={false} name={service.name} />
               )}
+            {service.isLoadingPage && !service.isFirstLoad && (
+                <WebviewLoader loaded={false} />
+            )}
             {service.isError && (
               <WebviewErrorHandler
                 name={service.recipe.name}
--- a/src/models/Service.js
+++ b/src/models/Service.js
@@ -74,6 +74,8 @@ export default class Service {
 
   @observable isLoading = true;
 
+  @observable isLoadingPage = true;
+
   @observable isError = false;
 
   @observable errorMessage = '';
@@ -383,11 +385,20 @@ export default class Service {
 
       this.hasCrashed = false;
       this.isLoading = true;
+      this.isLoadingPage = true;
       this.isError = false;
+    });
+
+    this.webview.addEventListener('did-stop-loading', event => {
+      debug('Did stop load', this.name, event);
+
+      this.isLoading = false;
+      this.isLoadingPage = false;
     });
 
     const didLoad = () => {
       this.isLoading = false;
+      this.isLoadingPage = false;
 
       if (!this.isError) {
         this.isFirstLoad = false;
@@ -407,6 +418,7 @@ export default class Service {
         this.isError = true;
         this.errorMessage = event.errorDescription;
         this.isLoading = false;
+        this.isLoadingPage = false;
       }
     });
 


vraravam (approved):

HenrijsS:
@SpecialAro What would you think about a more subtle Progressbar at the top?

Here's an easy library for it: https://kimmobrunfeldt.github.io/progressbar.js/

kris7t:
This might also partially fix https://github.com/ferdium/ferdium-app/issues/157

SpecialAro:
> This might also partially fix https://github.com/ferdium/ferdium-app/issues/157

For real? 🧐 Wasn't looking for that issue when coding, but I might have a hint on what's going wrong there... I think we might be handling event listeners wrongly, and that the variable `isLoading` doesn't turn to false when it should (stumbled across some problems, that's why I implemented another variable and added a new event listener as well).

kris7t:
well, it seems you also set `isLoading = false` in the `did-stop-loading` listener https://github.com/ferdium/ferdium-app/pull/248/files#diff-17d0da31b2d63aba19230c37f8f83212930e7493a7895464981c50897cfc4844R395 -- if that didn't mess up something else (like handling crashes during page loading or something), then the loader should have potentially became more robust. At least that's what I have observed when playing around with `BrowserView`: https://github.com/ferdium/ferdium-app/issues/157#issuecomment-1153215048 (which is a very different thing from from `<webview>`, so I'm not 100% certain the same fix could apply here).

But I still have no clue in what order should the various events fire on page loading...

OskarEichler:
@SpecialAro @vraravam To be honest I find this implementation incredibly annoying - now the Loading square is taking over every single time I type in Whatsapp or switch pages and is super distracting. Appreciate the thought of a loading indicator, but it should be less intrusive for sure! 

Maybe do something like a thin progress bar on the top or just a small loading indicator in the icon on the left side or something that is much more nuanced than a full page takeover in the middle of the screen please! 

Normal browsers like Chrome and Firefox don't do this either, so users are already used to the behavior of waiting for a short amount of time after interacting on a site. This feels like overkill. Or maybe make it an optional setting for us to turn off.

vraravam:
There's this https://www.npmjs.com/package/react-topbar-progress-indicator which can be less obstrusive. @SpecialAro - could you implement this? or shall i try later tonight after work?

SpecialAro:
> There's this https://www.npmjs.com/package/react-topbar-progress-indicator which can be less obstrusive. @SpecialAro - could you implement this? or shall i try later tonight after work?

Hey 👋
The next couple of days (today and tomorrow) will be difficult for me (as I have two weddings to be at) so if you can try I would really appreciate it!

I'll be available for anything you need on discord during the process, but won't be able to get my hands on a computer 🙈

pepper3k:
This is perfect!
Side note: I would suggest sticking with the ferdium color scheme (purple shade) so the appearance is consistent.

gustavotrott:
It would be good if I could disable it for a specific service!
I'm using WhatsApp and it is showing this Loading indicator very often and disrupting the usability.

vraravam:
@gustavotrott - the upcoming nightly (probably in an hour or so) will fix this.

OskarEichler:
@vraravam The progress bar is much better! Thanks for the swift implementation! Some feedback:

1. The progress bar is still very jittery for Whatsapp (maybe only display after a slight delay, so that when a lot of loads happening very quickly after one another it doesn't jitter like that and only starts displaying when a lot takes longer than 0.2 seconds or something along those lines)
2. Make the progress bar optional per service or add a global on/off toggle for users to decide if they want it or not
3. Right now the color is some rainbow gradient - would be cool to use the same color that was set in Settings > Appearance > Accent Color Settings

vraravam:
@OskarEichler :
1. Delayed + Merged action: The [npm package that I found](https://www.npmjs.com/package/react-topbar-progress-indicator), does not have this capability. By inferring what you want, you are also referring to the ability to "merge" multiple such load events into one, so that the progress-bar can suitably show completion status. Again, this is not present in the library.
2. Optional per service: I can try this out over this weekend - no promises though
3. If using the same as the accent color, then there's no visible difference between the progress bar and the window title (over which the progress bar is rendered). This will effectively make it useless. Instead, we would need to add another user-specifiable option which could then be injected for the progress bar configuration. More work, will try, no promises.

OskarEichler:
Sounds good, thank you @vraravam 

1. I'm more thinking that you only mount TopBarProgress with a delay of 0.2 seconds / if this.state.loading is still true after 0.2 seconds. There really is no need to show the progress bar for such short loading times and keep mounting and unmounting the component as it will just jitter onscreen. If you add a small delay function then it will only mount the component when it's actually a longer load, for which a loading indication is helpful. 

2. Awesome thank you!

3. Regarding accent color - on Mac there is no window title, so I'm not seeing that. Why not use a different shade of the primary color? In Accent Color Settings it's possible to set the primary color, and then choose between 5 different shades of brightness levels. Could just choose a lighter / darker version from the user setting so that there is a distinct difference, but it still fits the users settings and doesn't require a whole new color picker.

vraravam:
>     1. I'm more thinking that you only mount TopBarProgress with a delay of 0.2 seconds / if this.state.loading is still true after 0.2 seconds. There really is no need to show the progress bar for such short loading times and keep mounting and unmounting the component as it will just jitter onscreen. If you add a small delay function then it will only mount the component when it's actually a longer load, for which a loading indication is helpful.

I'm not a js/css dev, so won't be able to accomplish your recommendation. Will let one of the other contributors take a shot at this.

>     3. Regarding accent color - on Mac there is no window title, so I'm not seeing that. Why not use a different shade of the primary color? In Accent Color Settings it's possible to set the primary color, and then choose between 5 different shades of brightness levels. Could just choose a lighter / darker version from the user setting so that there is a distinct difference, but it still fits the users settings and doesn't require a whole new color picker.

But, here's the PR with a screenshot of the user-controllable progress bar: #291

gustavotrott:
Unfortunatelly the  less-obstrusive progress bar didn't help for WhatsApp.
The service become very slow because the hectic progress bar.

Once WhatsApp is the main service I use, I had to downgrade to 6.0.0-nightly.60 to be usable.

kris7t:
@gustavotrott You should be able to disable the progress bar per service as per https://github.com/ferdium/ferdium-app/pull/301

gustavotrott:
> @gustavotrott You should be able to disable the progress bar per service as per #301

Awesome! Thanks

Issue #185, opened 2022-05-21
Setup i18n with Crowdin


nathanaelhoun:
- [x] Create a Crowdin project : https://crowdin.com/project/ferdium-app
- [x] Setup the Crowdin → GitHub integration to submit daily PRs 
- [x] Fix the internationalization setup

The setup is a bit strange for i18n at the moment:
- The code languages listed in `src/i18n/locales` are not consistent: some are two characters locales code like `fr.json`, other are regionalized locale codes like `zh-HANT.json` or `en-US.json`
- The languages listed here in `src/i18n/languages.ts` (availables to the users) doesn't correspond to the languages listed in Crowdin, nor to the files present in the `src/i18n/locales/` folder. For instance:
    - `zh-HANT` is meant to be Chinese Traditional according to [some sources](https://www.alchemysoftware.com/livedocs/ezscript/Topics/Catalyst/Language.htm), but it's mapped to Chinese Simplified on Crowdin.
    - ~~"Nederlands" is marked in the UI, but it is "Dutch" on Crowdin.~~ expected as we write the native language names in the UI
    - Vietnamese is translated to 90% but is not available to the users. (fixed by #562)
    
We have decided to not hold the 6.0.0 release for this, as the current translations are working. But this will be fixed before allowing translations again through Crowdin.

If you see some strings that need to be fixed before 6.0.0, please report them below this issue.

comradekingu:
Some locales actually use country-specific codes.
However, your locale problem will not improve with Crowdin.
It creates non-workable alias locales, invalidating even good efforts.

Crowdin itself is a disservice to translators, as it is not possible to see more languages and draw from them, things get held up in voting, and large arrays of strings can't be browsed at once.
This makes for incomplete and inconsistent translations, at best reaching coverage by way of multiple translators having less oversight.

The "privacy" policy, has (among other things)
>behaviourally targeted advertising cookies - These cookies are used to deliver adverts more relevant to you and your interests. They are also used to limit the number of times you see an advertisement as well as help measure the effectiveness of the advertising campaigns. They are usually placed by advertising networks with the website operator’s permission. They remember that you have visited a website and this information is shared with other organisations such as advertisers. Quite often targeting or advertising cookies will be linked to site functionality provided by the other organisation.

pepper3k:
@comradekingu  Sounds to me like a good time to ditch this platform

comradekingu:
@eitanprezes Weblate is as viable as its differentiation from other platforms.
[...]

PR #284, opened 2022-06-17
fix: add Crowdin languages mapping in config file

andrii-bodnar:
Hey everyone,

just noticed there are some language codes in the project that have locale code instead of two letters code. Crowdin has a feature to map these codes.

Also, I've noticed that there a some duplicated languages in the locales folder - zh and zh-Hans, zh-Hant and zh-TW. I guess some of these languages are excessive here but not sure which. @nathanaelhoun could you please point out what codes are unneeded here? I will remove them also and will update the mapping if needed

#185


vraravam:
@nathanaelhoun - would you be able to verify and approve this change? IF so, we can then merge it in?

nathanaelhoun:
Hi @andrii-bodnar, thanks for taking a look, and sorry for the delay in my response!

Your comment about duplicated languages is correct. Some of them as excessive, but unfortunately, I do not know chinese well and I couldn't tell which one are duplicates. Do you know?

nathanaelhoun (approved):
Otherwise, these changes are looking good to me!

andrii-bodnar:
@nathanaelhoun the same, also not sure which of these files are excessive 😬

nathanaelhoun:
I'll just merge this one after the checks pass, and we will clear up the maybe excessive languages in another PR. Thanks @andrii-bodnar!

Issue #370, opened 2022-06-26
Bug: can't move window in mac


helloabella:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.0.0-beta.2

### What Operating System are you using?

macOS

### Operating System Version

macOS Monterey 12.4

### What arch are you using?

x64

### Last Known Working Ferdium version

6.0.0-nightly.69

### Expected Behavior

Able to drag app window around with the top section of the app

### Actual Behavior

Was able to drag window for a brief second before Skype fully loaded, but does not work on Whatsapp or Slack.

### Steps to reproduce

1. Open App
2. Attempt to drag window around

### Debug link

https://debug.ferdium.org/0b690e10-b6ad-4650-8181-5f250501d1e8

cino:
@helloabella I'm closing this issue as the problem has been resolved in #372 and released in the  v6.0.0-nightly.79 release. If you have any other issues feel free to open an issue again.

PR #348, opened 2022-06-24
Fix navigation toolbar buttons not being clickable (fixes #347)

vraravam:
#### Pre-flight Checklist

1. Please remember that if you are logging a bug for some service that has _stopped working_ or is _working incorrectly_, please log the bug [here](https://github.com/ferdium/ferdium-recipes/issues)
2. If you are requesting support for a **new service** in Ferdium, please log it [here](https://github.com/ferdium/ferdium-recipes/pulls)
3. Please remember to read the [self-help documentation](https://github.com/ferdium/ferdium-app#troubleshooting-recipes-self-help) - in case it helps you unblock yourself for issues related to older versions of recipes that were installed on your machine. (These will get automatically upgraded when you upgrade to the newer versions of Ferdium, but to get new recipes between Ferdium releases, this documentation is quite useful.)
4. Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

#### Description of Change

Moved the span for `window-draggable` into a conditional block and only include it in the DOM if the user has opted to show the draggeble area.

#### Motivation and Context

Fixes #347 

#### Screenshots

Jun-24-2022 07-36-00

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (`npm run prepare-code`)
- [x] `npm test` passes
- [x] I tested/previewed my changes locally

Issue #347, opened 2022-06-23
Bug: Toolbar not clickable in the horizontal area that OS level UI components plus padding use


BrianGilbert:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.0.0-nightly.74

### What Operating System are you using?

macOS

### Operating System Version

Monterey 12.4 (21F79)

### What arch are you using?

arm64 (including Apple Silicon)

### Last Known Working Ferdium version

unknown

### Expected Behavior

able to click in area of screen that service header UX is displayed

### Actual Behavior

Unable to click in this area

### Steps to reproduce

https://user-images.githubusercontent.com/114017/175430668-08fd3b3a-ed1d-4a95-86d7-d0a1a740d945.mp4

### Additional information

This affects all services

wgordon17:
I'm having a similar issue now. I can't move the window by clicking in the top area. Also on macOS Monterey (12.4).

This only happens if the Ferdium window is _already_ active. If it's not the active window, then I'm able to grab the top and move the window around. But after the window becomes active, I'm no longer able to move the window. This occurs both with and without the navigation bar.

https://user-images.githubusercontent.com/22222756/175781449-d858e023-0ff9-4439-a5ab-773724d72c56.mov

vraravam:
There was a bug where the draggable area on the top was present irrespective of whether the user had enabled it or not. The PR that fixed this issue had this fix in it as well. For you to enable the draggable area, you will have to turn it on in preferences first, and then you should be able to drag and drop.

wgordon17:
Perfect, I found it in the preferences, and this indeed brought back the functionality I was expecting. Thanks 👍

cino:
@wgordon17 You can check it again with that preference turned of the merged fix above enables dragging again without the extra bar. :) (Ofcourse,in the latest release)

PR #372, opened 2022-06-26
fix: bring the draggable area back on MacOS

cino:

#### Description of Change
This change brings the "invisible" draggable area back on macOS, the reason this was removed was that the elements behind it weren't clickable anymore. However, the reason this was happening wasn't actually because of the draggable area but because of the feature that enables the window to be maximized by double-clicking that area (hence the capturing of click events). 

My proposed solution is to bring the draggable area back and reduce the double-click area to 10px which increases the area where you can click and expect the normal behaviour.

#### Motivation and Context
- To improve UX and allow the window to be dragged while on macOS without visible draggable area

This should fix:
- #370 
- #347 
- While still keeping #348 happy as well

#### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/components/layout/AppLayout.js
+++ b/src/components/layout/AppLayout.js
@@ -65,7 +65,7 @@ const styles = theme => ({
     display: 'block',
     zIndex: 1,
     width: '100%',
-    height: '29px',
+    height: '10px',
     position: 'absolute',
     top: 0,
   },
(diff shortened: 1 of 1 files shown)


j7e:
Per the reply to #793 - https://github.com/ferdium/ferdium-app/issues/793#issuecomment-1324866173 - after exploring further, the Settings screen referred to in the reply is not directly accessible from the splash screen menu bar but only available once the user is past the splash screen, either by logging in or using without an account. The fix is not obvious to a new user; because Ferdium didn't behave in a standard way, my initial impression was that it was broken. At this point, a new user might reasonably conclude the same and abandon it, which does not seem desirable.

Could I suggest i) something on the splash screen to draw attention to troubleshooting issues (or this issue in particular) and ii) making the Settings screen available from the splash screen menu bar?

Issue #224, opened 2022-06-09
Your services have been updated - Reload services


leoheck:
### Your issue

Hello guys!

Yesterday I started to see this. This is pretty nice actually. 
![image](https://user-images.githubusercontent.com/1277920/172903359-ae776471-aac9-4010-a8b2-2da54751b291.png)

Every time this appears, I reload the services. But this thing keeps appearing. I am not sure if it is a bug or if these files a constantly being updated. It they are being updated it is fine, I will deal with it since it is a nightly build. But I would like to say that this thing is appearing quite often. However, it may be a bug where the service files are not being updated or something.

Version I am using, 
![image](https://user-images.githubusercontent.com/1277920/172903462-60f5936e-8824-490e-a4e7-81036f32ee47.png)

Cheers!

vraravam:
@leoheck - could you please try this with the latest nightly? We are already on 55+

vraravam:
Also - is this a duplicate of #204 ?

leoheck:
Sure, I was just not upgrading it since the Messages announced on this New Feed were never saying something related to my setup on Linux. They were mostly addressing things on Windows/macOS, but this may not be right. Maybe there are some features being hidden in this feed. 

See, it always shows things like "Other Changes" and maybe not the important changes. 

![image](https://user-images.githubusercontent.com/1277920/172989662-bf3f040e-2521-48af-938a-82ded5940d0f.png)

leoheck:
> Also - is this a duplicate of https://github.com/ferdium/ferdium-app/issues/204 ?

I am not sure. I am going to install the latest nightly now and give you some feedback by tomorrow.

vraravam:
We are using the autogenerated changelog and not manually calling out anything specific per se. That's why you might see only the section for "Other changes"

In the past couple of nightlies, we are have also upgraded electron to 19.0.4 (from 18.x) - so that is definitely an improvement across the board for all OSes - especially in the context of security fixes on chromium

leoheck:
Got it. Then I will update more often for sure. Thanks @vraravam !

SpecialAro:
> Also - is this a duplicate of #204 ?

Yes, this is a duplicate of that issue. I now why this is happening and still haven't got the time for raising an issue about it.

Basically, Ferdium checks the API server for a new "signal" on whether there is a recipe new version or not. And then, instead of downloading the new version from the API server, it checks the local instance for a version mismatch again, and triggers an install of that version (local) again.

[...]

PR #382, opened 2022-06-27
Disable API Server recipes updates

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
- Ferdium App won't install API Server recipes updates

#### Motivation and Context
As discussed in https://github.com/ferdium/ferdium-app/discussions/314 and https://github.com/ferdium/ferdium-app/issues/224, and also in discord with @vraravam, this PR disables **COMPLETELY** the in-app recipes updates from the server.

I think this is the way to go to have a fully functional release soon, by making the app totally independent of the recipes on the server.

Ergo, for the user to get recipes updates, it needs to download a new Ferdium version, which will - after the first run of the new updated version - pop up the Reload Services message to update recipes to the newest version (bundled with the App).

**DISCLAIMER: This can possible (not tested but most surely will) break the future of supporting custom recipes updates through a custom server - if users are using them to manage multiple account sync of custom recipes.**

#### Screenshots
Please, look at the console:

| Before <br /> image | After <br /> image |
|-- |-- |

#### Checklist

- [x] `npm test` passes

#### Release Notes

Disable API Server recipes updates

### DIFF
--- a/src/stores/RecipesStore.ts
+++ b/src/stores/RecipesStore.ts
@@ -94,8 +94,9 @@ export default class RecipesStore extends TypedStore {
 
     if (Object.keys(recipes).length === 0) return;
 
-    const remoteUpdates = await this.getRecipeUpdatesRequest.execute(recipes)
-      ._promise;
+    // TODO: This line needs to be uncomment once we fix the App-Server interaction problem.
+    // const remoteUpdates = await this.getRecipeUpdatesRequest.execute(recipes)._promise;
+    const remoteUpdates = [];
 
     // Check for local updates
     const allJsonFile = asarRecipesPath('all.json');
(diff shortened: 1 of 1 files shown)


src/stores/RecipesStore.ts
```diff
@@ -113,7 +113,10 @@ export default class RecipesStore extends TypedStore {
       }
     }
 
-    const updates = [...remoteUpdates, ...localUpdates];
+    const updates = [
+      // ...remoteUpdates, //TODO: This line needs to be uncomment once we fix the App-Server interaction problem.
```
vraravam: Instead of commenting this line, can we instead comment line 97? That way, the network call also will not happen. We can then initialize the `remoteUpdates` const as an empty array `[]` - so that all other lines can remain as-is.

vraravam (changes requested):

src/stores/RecipesStore.ts
SpecialAro: Yes, it is better that way!

vraravam (approved):

PR #432, opened 2022-07-06
Checks if MSVS 2019 or 2022 exists through reg key

SlanDF:
#### Pre-flight Checklist

#### Description of Change

Ive added some changes to the .ps1 script that the system dependencies are at the correct version. In this case I've made the validation for the MSVC to be made using the path inside the registry to make sure we have the correct path no matter where MSVS is installed. I fetch the directory from there then travel to that directory, pop the path once and that places you in the folder that contains the version of the MSVST that you want. Then it checks if the 2019 or 2022 folders are there, if they are then it passes the validation. I kept the same error handling that was there before, and only commented the previous code that handled this part of the script. I also added error handling in case the registry path to vs doesnt exist. 

#### Motivation and Context

The way it was previously implemented didnt really made much sense, and i was having issue passing that validation knowing fully well i had my MSVST updated. So I went and digged on how to find this version you wanted keeping in consideration it can be installed anywhere (hence the reg key solution)
I honestly think this could even be changed to go by version number because thats was more easy to access in the registry and you could just set if the version>14 or something

#### Checklist

- [x] `npm test` passes

#### Release Notes

none

### DIFF
--- a/scripts/build-windows.ps1
+++ b/scripts/build-windows.ps1
@@ -115,15 +115,6 @@ if(-not $ACTUAL_MSVST_VERSION){
         expected: [$EXPECTED_MSVST_VERSION]"
 }
 
-# -------------------------LEGACY----------------------------------
-# Ensure that the system dependencies are at the correct version - fail if not
-# Check MSVS Tools through MSVS_VERSION
-# if (-not ($EXPECTED_MSVST_VERSION -contains $ACTUAL_MSVST_VERSION)) {
-#   fail_with_docs "You are not running the expected version of MSVS Tools!
-#     expected: [$EXPECTED_MSVST_VERSION]
-#     actual  : [$ACTUAL_MSVST_VERSION]"
-# }
-
 # -----------------------------------------------------------------------------
 # Ensure that the system dependencies are at the correct version - recover if not
 # Check npm version
--- a/scripts/build-windows.ps1
+++ b/scripts/build-windows.ps1
@@ -40,7 +40,6 @@ Function Test-CommandExists { Param ($command, $1)
 # Check for installed programmes
 Test-CommandExists node "Node is not installed"
 Test-CommandExists npm "npm is not installed"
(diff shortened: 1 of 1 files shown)


scripts/build-windows.ps1
```diff
@@ -94,16 +93,37 @@ if ($env:CLEAN -eq "true")
...
+}
+
+if(-not $ACTUAL_MSVST_VERSION){
   fail_with_docs "You are not running the expected version of MSVS Tools!
-    expected: [$EXPECTED_MSVST_VERSION]
-    actual  : [$ACTUAL_MSVST_VERSION]"
+        expected: [$EXPECTED_MSVST_VERSION]"
 }
 
+# -------------------------LEGACY----------------------------------
```
vraravam: Please remove this commented out portion. If needed, we can revert the commit to resurrect it

vraravam (changes requested):

vraravam:
Also - in the new commit that you are going to do, please add `[skip ci]` at the end of the commit message.
Reasoning: This change has no direct impact on the software being built. Moreover, its only for local builds on contributors machines - and thus we can avoid triggering a CI build as its unnecessary.

SpecialAro:
image
@Kanubiz this is currently failing on my system. I can assure it was well on the other script.

vraravam:
@SpecialAro - what's the version that's printed if you run that line manually in your system?

SlanDF:
Me and @SpecialAro figured a better way to check the actual version
`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64` Includes the version number that you want to really check
v14.20 to v14.29 is 2019
v14.30 to v14.32 (latest) is 2022
like this we just need to go to that reg path then assign ` $ACTUAL_MSVST_VERSION` to either 2019 or 2022 based on the above conditions

PR #437, opened 2022-07-06
Add file size info and error for import custom image on recipe

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
This adds the ability of having the file size info and error on the import image component

#### Motivation and Context
When you try to import a file larger than 1 MB, ~~an error of 413 (nginx) happens on the server side (if you're using Ferdium Server)~~ (changed to 2MB on nginx side). This ensures that the user imports a smaller filer size.
If you're using Ferdium without an account this value is ALSO 2 MB ~~altough this PR only accounts for the above situation (maybe we can also change it on Ferdium Server nginx side to ensure the same value)~~

#### Screenshots
| Before <br /> image | After <br /> image |
|-- |-- |

#### Checklist

- [x] `npm test` passes

#### Release Notes

Add file size info and error for import custom image on recipe

### DIFF
--- a/src/components/settings/services/EditServiceForm.js
+++ b/src/components/settings/services/EditServiceForm.js
@@ -406,7 +406,7 @@ class EditServiceForm extends Component {
                   field={form.$('customIcon')}
                   textDelete={intl.formatMessage(messages.iconDelete)}
                   textUpload={intl.formatMessage(messages.iconUpload)}
-                  maxSize={1_048_576}
+                  maxSize={2_097_152}
                   maxFiles={1}
                   textMaxFileSize={intl.formatMessage(messages.maxFileSize)}
                   textMaxFileSizeError={intl.formatMessage(
--- a/src/components/settings/services/EditServiceForm.js
+++ b/src/components/settings/services/EditServiceForm.js
@@ -141,6 +141,14 @@ const messages = defineMessages({
     id: 'settings.service.reloadRequired',
     defaultMessage: 'Changes require reload of the service',
   },
+  maxFileSize: {
+    id: 'settings.service.form.maxFileSize',
+    defaultMessage: 'Maximum filesize:',
+  },
+  maxFileSizeError: {
+    id: 'settings.service.form.maxFileSizeError',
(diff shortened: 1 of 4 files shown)


src/components/ui/ImageUpload.tsx
```diff
@@ -40,14 +54,43 @@ class ImageUpload extends Component<Props> {
     field.set('');
   }
 
+  onDropRejected(rejectedFiles) {
```
cino:
```suggestion
  onDropRejected(rejectedFiles): void {
```

Also, could you add the type of the rejectedFiles, I'm not sure what it is tho

src/components/ui/ImageUpload.tsx
```diff
@@ -40,14 +54,43 @@ class ImageUpload extends Component<Props> {
...
+      maxSize,
+      maxFiles,
+    } = this.props;
 
     const cssClasses = classnames({
       'image-upload__dropzone': true,
       [`${className}`]: className,
     });
 
+    const maxSizeParse =
```
cino:
```suggestion
    const maxSizeParse: number =
```

This will make extra sure that maxSize is a number as well

src/components/settings/services/EditServiceForm.js
```diff
@@ -398,6 +406,12 @@ class EditServiceForm extends Component {
                   field={form.$('customIcon')}
                   textDelete={intl.formatMessage(messages.iconDelete)}
                   textUpload={intl.formatMessage(messages.iconUpload)}
+                  maxSize={2_097_152}
```
vraravam, line 409: Does it make sense to hardcode this number or use `Number.POSITIVE_INFINITY`?

SpecialAro: The function by default uses it as infinity, so I thought we should parse the same. Tho, maybe we can change that if it's unwanted.

The limit is hardcoded there because it's component specific (as we might have different sizes for different components) if we add it somewhere else.

vraravam (approved):
LGTM!

PR #439, opened 2022-07-06
refactor: refactored MSVS validation

SlanDF:
#### Pre-flight Checklist

#### Description of Change

Now checking by version in the registry then picking 2022, 2019 or 2017 or lower depending on the version for better control, and then `npm config set msvs_version $ACTUAL_MSVST_VERSION
`
#### Motivation and Context

It wouldnt work properly if you had more than one version installed, and it wasnt setting it in the npm config
#### Checklist

- [x] `npm test` passes

### DIFF
--- a/scripts/build-windows.ps1
+++ b/scripts/build-windows.ps1
@@ -94,31 +94,40 @@ if ($env:CLEAN -eq "true")
 # -----------------------------------------------------------------------------
 # Ensure that the system dependencies are at the correct version - fail if not
 # Check MSVS Tools through MSVS_VERSION
+
+
 $EXPECTED_MSVST_VERSION = @("2019","2022")
-# TODO: Implement path for ARM machines
-$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64"
-
-if(-not (Test-Path -Path $MSVS_REG_PATH)){
-  fail_with_docs "You don't have the Microsoft Visual Studio Tools installed!"
-}
-
-$MSVS_VERSION =  [int]((Get-ItemProperty -Path $MSVS_REG_PATH).Version.substring(4, 2))
-switch($MSVS_VERSION) {
-  { $MSVS_VERSION -ge 30 } {$ACTUAL_MSVST_VERSION = "2022"}
-  { ($MSVS_VERSION -ge 20) -and ($MSVS_VERSION -le 29) } {$ACTUAL_MSVST_VERSION = "2019"}
-  { $MSVS_VERSION -lt 20 } {$ACTUAL_MSVST_VERSION = "2017 or lower"}
-}
-
-if (-not ($EXPECTED_MSVST_VERSION -contains $ACTUAL_MSVST_VERSION)) {
-  fail_with_docs "You are not running the expected version of MSVS Tools!
-    expected: [$EXPECTED_MSVST_VERSION]
-    actual  : [$ACTUAL_MSVST_VERSION]"
-}
-
 $NPM_CONFIG_MSVS_VERSION = npm config get msvs_version
 if((-not $NPM_CONFIG_MSVS_VERSION) -or -not ($EXPECTED_MSVST_VERSION -contains $NPM_CONFIG_MSVS_VERSION)){
+  Write-Host "Your Microsoft Visual Studio Tools isn't set properly or it's not the right version!
+              Checking your version..."
+
+  # TODO: Implement path for ARM machines
+  $MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64"
+
+  if(-not (Test-Path -Path $MSVS_REG_PATH)){
+    fail_with_docs "You don't have the Microsoft Visual Studio Tools installed!"
+  }
+
+  $MSVS_VERSION =  [int]((Get-ItemProperty -Path $MSVS_REG_PATH).Version.substring(4, 2))
+  switch($MSVS_VERSION) {
+    { $MSVS_VERSION -ge 30 } {$ACTUAL_MSVST_VERSION = "2022"}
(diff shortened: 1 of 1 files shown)


scripts/build-windows.ps1
```diff
@@ -95,25 +95,26 @@ if ($env:CLEAN -eq "true")
 # Ensure that the system dependencies are at the correct version - fail if not
 # Check MSVS Tools through MSVS_VERSION
 $EXPECTED_MSVST_VERSION = @("2019","2022")
-$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\Setup"
+
+$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64"
 
 if(-not (Test-Path -Path $MSVS_REG_PATH)){
-  fail_with_docs "You don't have MSVS Tools Installed!"
+  fail_with_docs "You dont have the Microsoft Visual Studio Tools installed!"
```
cino:
```suggestion
  fail_with_docs "You don't have the Microsoft Visual Studio Tools installed!"
```

scripts/build-windows.ps1
```diff
@@ -95,25 +95,26 @@ if ($env:CLEAN -eq "true")
 # Ensure that the system dependencies are at the correct version - fail if not
 # Check MSVS Tools through MSVS_VERSION
 $EXPECTED_MSVST_VERSION = @("2019","2022")
-$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\Setup"
+
+$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64"
```
vraravam: This last part of the path might not work for arm machines.

vraravam (changes requested):

scripts/build-windows.ps1
SlanDF: how can we test it?

vraravam: maybe ask if some of the contributors has an arm machine running windows?

vraravam: or maybe we just put a `TODO` comment stating this (above line 99) and merge this PR. If/when we get someone with an arm machine AND using windows, then we can revisit?

SpecialAro: Also, is this version specific? Does this work for 2019 and 2022 build tools ?

SlanDF: well, im sure someone with and ARM machine if he tries to build, hell try to come as ask about it and we can see how the path is different

vraravam (approved):

scripts/build-windows.ps1
```diff
@@ -95,25 +95,26 @@ if ($env:CLEAN -eq "true")
...
 }
 
-if(-not $ACTUAL_MSVST_VERSION){
+if (-not ($EXPECTED_MSVST_VERSION -contains $ACTUAL_MSVST_VERSION)) {
   fail_with_docs "You are not running the expected version of MSVS Tools!
-        expected: [$EXPECTED_MSVST_VERSION]"
+    expected: [$EXPECTED_MSVST_VERSION]
+    actual  : [$ACTUAL_MSVST_VERSION]"
 }
+npm config set msvs_version $ACTUAL_MSVST_VERSION
```
SpecialAro: I don't think we should be doing this. Was the user prompt previously?

In my opinion we should only set this if the user is prompt... For instance: I have two instances of msvs installed and for ferdium I use 2022... I should be able to select it

I think we can check if it's set with `npm config get` before setting it, and only set it if the user doesn't have it configured.

SpecialAro (changes requested):

scripts/build-windows.ps1
```diff
@@ -95,25 +95,26 @@ if ($env:CLEAN -eq "true")
 # Ensure that the system dependencies are at the correct version - fail if not
 # Check MSVS Tools through MSVS_VERSION
 $EXPECTED_MSVST_VERSION = @("2019","2022")
-$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\Setup"
+# TODO: Implement path for ARM machines
+$MSVS_REG_PATH = "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64"
```
SpecialAro: Also, is this version specific? Does this work for 2019 and 2022 build tools ?

scripts/build-windows.ps1
SlanDF: yeah we can get before setting for sure, the switch case makes it so it fetches the highest version but im guessing that would be personal preference if you got multiple installed right

SpecialAro: > personal preference if you got multiple installed right

Yes... but the problem is that if you have different projects that use node and different msvs versions this will mess up with your config.
So, my suggestion is to check if msvs_version is not set, and then set it. If it is set to a version that is not supported, add a message to the user and break the script.

SpecialAro (changes requested):

scripts/build-windows.ps1
SlanDF: @SpecialAro it should yes, if you want you can confirm, but when i had both 2019 and 2022, the only reg key is the highest aka 2022
the 14.0 on the path is all the versions from 2015 upwards
the check wether its 2022, 2019, 2017 or lower is in the switch case

SlanDF: if you manage to find me a case where you got 2 registry values then we gotta look into it but at least on my pc once i installed 2022 it overrided the previous one

SpecialAro (approved):

vraravam (approved):
I assume this is working - i dont have windows. but, the logic seems correct.

scripts/build-windows.ps1
vraravam: I would suggest that you try out `direnv`: as you cd into the folder, it will set this config, and then when you cd out, it will remove it. I think it works in git-bash or cygwin in windows.

PR #459, opened 2022-07-09
Remove msi builds on windows

Alphrag:
#### Pre-flight Checklist

#### Description of Change

Delete the section that was building the `msi` installer on windows.

#### Motivation and Context

It has appeared recently that the `msi` builds have an inconsistent behaviour with the rest of the installers, meaning that the app doesn't appear with its correct version in the Windows Control Panel, and this behaviour can affect later updates that do not recognise the version is higher. Since they also do not have arm installer, it is preferable to use the full installer instead which contains all architectures.

#### Checklist

#### Release Notes

Note: Remove the `msi` installers for Windows which caused problems

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -77,8 +77,6 @@ win:
   target:
     - target: nsis
       arch: [x64, ia32, arm64]
-    - target: msi
-      arch: [x64, ia32]
     - target: portable
       arch: [x64, ia32, arm64]
   # The name of the CN appearing in the certificate must be present in the publisherName list below
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -15,7 +15,7 @@ snap:
 nsis:
   perMachine: false
   oneClick: true
-  artifactName: "${productName}-${os}-AutoSetup-${version}.${ext}"
+  artifactName: "${productName}-${os}-AutoSetup-${version}-${arch}.${ext}"
 
 portable:
   artifactName: "${productName}-${os}-Portable-${version}-${arch}.${ext}"


SpecialAro (commented):
If we want to pursue this, maybe we should announce this change... This will break usability for users with the current msi version, am I correct? They won't be able to update without manually downloading and configure with the exe installer, right?

Should we try to make some counter measures for that before merging this PR?

(Nevertheless, my personal opinion is that this PR is really important, and we should merge it and get rid of the msi package entirely - at least until we find a way to fix inconsistencies).

Alphrag:
tldr: This won't break anything. All updates will use the `exe` without the user noticing it (until they look at the control panel). So we can get rid of it without further thoughts.

Explanation: the way the auto-update works (for any OS) is that it fetches the file called `latest*.yml` on the release and use the url present in there. This means that for Windows, the auto-update feature *always* uses the `AutoSetup*.exe` file and never the `msi` one (on linux it is also the reason why only AppImages can auto-update). 
This is one of the other things showing that the way electron-builder works with the `msi` assets is problematic since it will be replaced by the operation of the `exe` installer at the first update. Currently, if you install through `msi` it creates a wrong entry in the Control Center, and then updating to the most up to date version creates another entry which now holds the correct informations. This should not happen! [I am planning to fill an issue on `electron-builder` directly as soon as I have time]

Alphrag:
As we remove them, would it be worth exposing the `AutoSetup` installers for the different architecture separately, since the combined one is fairly heavy? The auto-updater will choose the one closer to the arch of the user, and thus save bandwidth. What do you think?

vraravam (approved):

vraravam:
> As we remove them, would it be worth exposing the `AutoSetup` installers for the different architecture separately, since the combined one is fairly heavy? The auto-updater will choose the one closer to the arch of the user, and thus save bandwidth. What do you think?

If there's a simple way to do that, lets go for it. But, I would vote for whatever can unblock `6.0.0`

vraravam:
@SpecialAro @Alphrag  - Should we merge this - and test out the effect using the next nightly?

Alphrag:
> If there's a simple way to do that, lets go for it. But, I would vote for whatever can unblock `6.0.0`

Yep, the simple way is what I've just pushed: by forcing the artifact name to use the architecture, it separates all the installers.

SpecialAro:
Thank you @Alphrag for your detailed explanation 😁

Then, we should go for it! Let's merge this PR

SpecialAro (approved):

Issue #4, opened 2022-04-16
Switching from Ferdi to Ferdium


vraravam:
With the merge of PR #2 , we now have a kind of broken system. Pre-existing users of Ferdi would have had their profiles in different location (for eg, in macos, the ferdi profile folder is: `~/Library/Application Support/Ferdi`. With the new name, it will be `~/Library/Application Support/Ferdium`. We might need to either have a page documenting how pre-existing users can move over without losing any data (just needs renaming of this folder) or provide an automated way to move them over. Just logging this bug ticket so that someone can take this up prior to our releasing any versions.

lheinke:
I tried to migrate from Ferdi to Ferdium on my computer, but was not able to. Since the migration script is for Windows only, I simply copied the contents of `~/.config/Ferdi` to `~/.config/Ferdium` but this did not work. Any helpful input on what could be the problem? My Ferdi instance is quite old, so the config folder takes up  almost 7 GB.

vraravam:
Thanks for trying out Ferdium @lheinke !

If all you wanted to do was to migrate your local config folder from ferdi to ferdium, please check if these steps work:
1. Open Ferdi.
2. Go to `Settings > Advanced` and click on `Open Profile folder`. This should open the file explorer of your OS.
<img width="901" alt="Screenshot 2022-04-28 at 5 20 15 AM" src="https://user-images.githubusercontent.com/69629/165732296-3dc9921d-ad35-4bf9-bacd-1ee452f02c94.png">

3. Close Ferdi
4. Back in the file explorer of your OS, rename the profile folder from `Ferdi` to `Ferdium`
5. Start Ferdium

This should now start with your exact configurations from Ferdi.
Note: If you have any custom services, (not from the in-built services in Ferdium), you might need to edit them manually.

Please confirm if these steps help.

If you are trying to move away from the hosted server that holds your config/preferences, then please use the [following guide](https://github.com/ferdium/ferdi/blob/main/MIGRATION.md)

krisrok:
I also copied manually from `%appdata%\Ferdi` to `%appdata%\Ferdium` and it also did not work for me. Config is being loaded (at least my GUI settings from Ferdi) but Ferdium only says "Could not load services and user information".

I was able to add my services by exporting my serverless profile with Ferdi and importing with Ferdium (using the web interface on http://localhost:45569). But then my GUI config was left behind. I did not find a way to make it both work.

lheinke:
[...]

PR #473, opened 2022-07-12
Handle both ferdi and ferdium servers for transferring from

vraravam:
#### Pre-flight Checklist

#### Description of Change

When checking in pre-existing profiles whether Ferdium is being used without a server, we need to handle both Ferdi as well as Ferdium profiles.

#### Motivation and Context

Fixes comment from: https://github.com/ferdium/ferdium-app/issues/4#issuecomment-1181701753

#### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/api/apiBase.ts
+++ b/src/api/apiBase.ts
@@ -53,14 +53,18 @@ export function importExportURL() {
 
 export function serverBase() {
   const serverType = (window as any).ferdium.stores.settings.all.app.server;
-  const noServer = 'You are using Ferdium without a server';
+  const noServerFerdi = 'You are using Ferdi without a server';
+  const noServerFerdium = 'You are using Ferdium without a server';
 
   let terms;
   switch (serverType) {
     case LIVE_FRANZ_API:
       terms = DEV_API_FRANZ_WEBSITE;
       break;
-    case noServer:
+    case noServerFerdi:
+      terms = LIVE_FERDIUM_API;
+      break;
+    case noServerFerdium:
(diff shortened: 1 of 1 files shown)


SpecialAro (approved):
Looks good, but I'm not exactly sure if it will suffice to solve the issue. Have you test it?

real-or-random:
Hi, thanks for the quick attempt to fix this. I tested this with `v6.0.0-nightly.100` and it does not work. 

---

A quick code search shows that there is actually another code location where the magic string appears:
https://github.com/ferdium/ferdium-app/blob/d3f2edf4803d0a163b614a5a23842106b0068748/src/config.ts#L185

This is used a few places (and should also been used in code that has been touched by this PR). 

This location looks like a potential suspect:
https://github.com/ferdium/ferdium-app/blob/71c52373f81cace664047edd19d9d289f45a4dff/src/stores/SettingsStore.ts#L59-L62 

There's a `_migrate()` function that's called before that code is reached. I suspect the right thing to do is to add code to this function that rewrites the old string to the new string. Or better, migrate all "You are using ... " some better than a magic string that contains the appname, maybe just the empty string.

If I change the string manually in the `settings.json`, then everything works. So apparently that setting is really the only culprint. So this should be a reliable way for you to test this. Or alternatively, this also suggests a simpler solution: add a line to the migration shell scripts that rewrites the config...

vraravam:
Good catch and analysis @real-or-random - i will try to fix this later tonight

PR #482, opened 2022-07-13
fix: horizontal sidebar style on macOS

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
- Fix the horizontal sidebar style on macOS
- Fix the var name for handling horizontal sidebar (it was called `useVerticalStyle` because when I implemented it I was dumb... 👎 )
- Fix lint in some files
- Add correct hamburger menu icon in vertical styling

#### Motivation and Context
It was looking bad on the horizontal sidebar.
Works for now, but all these styles are pretty badly implemented, we should make a reform on it soon...

Had to revert the previous implementation. There is a need for a draggable space on the top - elements placed there won't get correctly animated (before clicking add recipe the cursor should turn to a hand instead of a pointer). If we want to narrow that space in the future this needs fixing.

#### Screenshots
| Before | After |
|-- |-- |
| image image image | image image image |

#### Checklist

- [x] `npm test` passes

#### Release Notes

fix: horizontal sidebar style on macOS


SpecialAro:
Found an issue with the buttons on the side bar. Don't merge until it's fixed please.

vraravam (approved):

vraravam:
Also - please verify when the sidebar is in vertical mode

src/features/appearance/index.ts
```diff
@@ -247,7 +249,17 @@ function generateServiceRibbonWidthStyle(
...
+    }
+    .darwin .app .app__content {
+      padding-top: ${
+        width + sidebarSizeBias + (sizeDragArea !== 0 ? 4 - sizeDragArea : 4)
+      }px !important;
+    }
+    .darwin .workspaces-drawer {
+      margin-top: -${
+        width + verticalStyleOffset - 5 - sizeDragArea
+      }px !important;
```
SpecialAro, line 262: This might be weird but it's the pixels it needs to align properly. In the future, we should fully move to a fully functional flexbox implementation for this.

src/config.ts
```diff
@@ -261,7 +261,7 @@ export const DEFAULT_APP_SETTINGS = {
   iconSize: iconSizeBias,
   navigationBarBehaviour: 'custom',
   searchEngine: SEARCH_ENGINE_STARTPAGE,
-  useVerticalStyle: false,
+  useHorizontalStyle: false,
```
vraravam, line 264: This will mean a breaking change for all pre-existing users - are we sure we want to rename? imo, its still ok to do so - coz "Ferdium" is still not released - but, we need to capture this in a release note.

SpecialAro: > This will mean a breaking change for all pre-existing users - are we sure we want to rename? imo, its still ok to do so - coz "Ferdium" is still not released - but, we need to capture this in a release note.

Yep, it is. I can revert it and leave it the other way around if needed. But we should change the code accordingly so that it makes sense.

Nevertheless, the change users will experience is that they need to toggle the toggle back again (I think that's it).

vraravam: yes - please call that out in the release notes when this gets released.

PR #491, opened 2022-07-17
Feature: Add Release Notes

### LINKED ISSUE #487: Feature: Show release / changelog notes prior to triggering an update if a user clicks the update option when a new update is available
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Currently it seems that with "Enable updates" and "Include pre-releases" enabled that it is very easy to accidentally be unaware of a change, problem or something else.

### Proposed Solution

If there were a method to show the release notes / changelog prior to the user accepting to actually install an update, that might be good insomuch that it will act as a warning and/or make users aware of what to expect (such as [the last update](https://github.com/ferdium/ferdium-app/releases/tag/v6.0.0-nightly.100) that included the horizontal bar change and requirement to toggle off and on).

Another thing that may be useful would be to, instead of - with PRE-RELEASE ONLY - triggering an auto-update, perhaps require the user to click the "Check for updates" function (eg, on macOS under menubar: `Ferdium` → `Check for updates`) to use more awareness and thoughtfulness before blasting in an update...

### Alternatives Considered

Another choice, that wouldn't get as much testing for the (auto)update system would be to just force manual updates for pre-release vs allowing users to just trigger it. 

Yet another choice would be, in the same manual line as above, when they click auto-update function on the bottom popup updater bar, it would lead them over to the releases page to be forced to have the release notes more readily available.

### Additional Information

_No response_

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Feature: Add Release Notes to Auth screen, Menu and Settings (New tab and Update section)

#### Motivation and Context
This fixes https://github.com/ferdium/ferdium-app/issues/487.
Had to implement a new component for Auth screen (with a new route too) because the original component was build on top of the SettingsScreen component - some props aren't available on the Auth screen. If we have a better idea how to implement this in the future, we can make refactor at that moment.

Please, test this on all OSes before merging - for testing, please change the app version manually on `package.json` to a lower version than the current and build it (make sure to run `npm i` given that two new dependencies need to be installed). Make sure to build for production, given that in dev mode the AutoUpdate feature doesn't work properly (love you electron <3)

Also, if you see some cleanup that can be done before merging, please do push a commit to clean things up.

Tested so far:
- [x] Windows
- [x] macOS
- [x] Linux

#### Screenshots
build-release-notes

#### Checklist

- [x] `npm test` passes

#### Release Notes

Feature: Add Release Notes


src/components/layout/AppLayout.jsx
```diff
@@ -126,86 +132,91 @@ class AppLayout extends Component {
...
+                    </InfoBar>
+                  )}
+                {automaticUpdates &&
+                  appUpdateIsDownloaded &&
+                  this.state.shouldShowAppUpdateInfoBar && (
+                    <AppUpdateInfoBar
+                      onInstallUpdate={installAppUpdate}
+                      updateVersionParsed={updateVersionParse(updateVersion)}
+                      onHide={() => {
+                        // eslint-disable-next-line react/no-unused-state
```
vraravam: i think this might mean some lurking issue - can the issue not be fixed without disabling eslint?

src/components/settings/releaseNotes/ReleaseNotesDashboard.jsx
```diff
@@ -0,0 +1,110 @@
...
+  handleClick(e) {
+    e.preventDefault();
+    openExternalUrl(e.target.href);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener(
+      'click',
+      // eslint-disable-next-line unicorn/no-invalid-remove-event-listener
+      this.handleClick.bind(this),
```
vraravam: Is this binding on unmounting?

src/components/settings/releaseNotes/ReleaseNotesLayout.jsx
```diff
@@ -0,0 +1,77 @@
...
+    document.removeEventListener(
+      'keydown',
+      // eslint-disable-next-line unicorn/no-invalid-remove-event-listener
+      this.handleKeyDown.bind(this),
+      false,
+    );
+  }
+
+  handleKeyDown(e) {
+    if (e.keyCode === 27) {
```
vraravam: please assign the `27` to a meaningful constant

src/helpers/update-helpers.ts
```diff
@@ -0,0 +1,14 @@
+import { ferdiumVersion } from '../environment-remote';
```
vraravam: These functions might be better in the `environment-remote` file? Just checking which is the better location. Also, have you added tests for these?

src/lib/Menu.js
```diff
@@ -586,10 +582,9 @@ const _titleBarTemplateFactory = (intl, locked) => [
       {
         label: intl.formatMessage(menuItems.changelog),
         click() {
-          openExternalUrl(
-            `${GITHUB_FERDIUM_URL}/ferdium-app/releases/tag/v${ferdiumVersion}`,
-            true,
-          );
+          window.location.href = window.location.href.includes('#/auth')
```
vraravam: This block of code/logic seems duplicated - can we move them into a utility class?

vraravam (changes requested):

cino (commented):
General comment: could you make sure that all **_new_** files are either ts or tsx :)

src/components/layout/AppLayout.jsx
SpecialAro: Yep, disabling this didn't cause any issue.

src/components/settings/releaseNotes/ReleaseNotesDashboard.jsx
SpecialAro: Yes, I followed this: https://github.com/ferdium/ferdium-app/blob/b85efe342e727905eb227bf227a9d1f06ae48e79/src/components/settings/SettingsLayout.jsx#L29-L36

Should we remove it?

src/components/settings/releaseNotes/ReleaseNotesLayout.jsx
SpecialAro: Also followed this: https://github.com/ferdium/ferdium-app/blob/b85efe342e727905eb227bf227a9d1f06ae48e79/src/components/settings/SettingsLayout.jsx#L38-L43

Should we assign a global constant for this?

src/helpers/update-helpers.ts
SpecialAro: Haven't added tests for this (actually don't know how to do that xD).

I'm moving it to `environment-remote`, then.

cino: @SpecialAro This should be quite simple to test actually (you can look at https://github.com/ferdium/ferdium-app/blob/develop/test/helpers/url-helpers.test.ts for example tests)

The only thing you might need to change a bit is getFerdiumVersion() and pass the `window.location.href` as a property, that way the function itself doesn't need any outside information and makes it unit testable. Hit me up on Discord if you need any help :)

SpecialAro:
Thank you @cino for your incredible help with this PR :D

src/components/settings/releaseNotes/ReleaseNotesDashboard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      'An error occured when connecting to Github, the page you are looking for is missing.',
+  },
+});
+
+interface IProps {
+  intl: any;
+}
+
+class ReleaseNotesDashboard extends Component<IProps> {
+  static propTypes = {};
```
vraravam: do we still need `propTypes`?

vraravam (approved):
LGTM!

src/containers/settings/ReleaseNotesWindow.tsx
```diff
@@ -0,0 +1,44 @@
...
+    this.portalRoot.append(this.el);
+  }
+
+  componentWillUnmount(): void {
+    this.el.remove();
+  }
+
+  render(): ReactPortal {
+    // const { closeSettings } = this.props.actions.ui;
+
```
cino:
```suggestion
```

cino (approved):

PR #532, opened 2022-08-01
chore: change values with mobx actions

### LINKED ISSUE #438: Tech: Stop changing MOBX observables without using an action
### Your issue

By updating Mobx/mobx-router to the latest versions there a lot of new warnings are introduced within the developer console, this can be disabled by configuring the following:

```ts
import { configure } from "mobx"

configure({
    enforceActions: "never",
})
```

However: I would like to keep this issue here until we actively stop modifying the state in such a manner, this doesn't seem to affect the application at this point but it would be good practice.

Example of warnings:
image

(ignore the error)

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Change values with mobx actions.
Cleanup the console warnings about mobx.

#### Motivation and Context
This fixes https://github.com/ferdium/ferdium-app/issues/438. I don't know if this is the proper way to fix this issue, but at least now the warnings are gone 😄 

May I ask people to test this before merging to see if it doesn't break any functionality? Thank you 😃 

#### Screenshots
| Before | After |
|-- |-- |
| image <br /> and plenty more... | image |

#### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/features/serviceProxy/index.ts
+++ b/src/features/serviceProxy/index.ts
@@ -1,4 +1,4 @@
-import { autorun, observable } from 'mobx';
+import { autorun, action, observable } from 'mobx';
 import { session } from '@electron/remote';
 
 const debug = require('../../preload-safe-debug')(
@@ -15,8 +15,12 @@ export default function init(stores: {
 }) {
   debug('Initializing `serviceProxy` feature');
 
+  const setIsEnabled = action((value: boolean) => {
+    config.isEnabled = value;
+  });
+
   autorun(() => {
-    config.isEnabled = true;
+    setIsEnabled(true);
 
     const services = stores.services.enabled;
     const proxySettings = stores.settings.proxy;
--- a/src/features/webControls/containers/WebControlsScreen.jsx
+++ b/src/features/webControls/containers/WebControlsScreen.jsx
@@ -2,7 +2,7 @@ import { Component } from 'react';
 import { observer, inject } from 'mobx-react';
 import PropTypes from 'prop-types';
 
-import { autorun, makeObservable, observable } from 'mobx';
+import { autorun, action, makeObservable, observable } from 'mobx';
 import WebControls from '../components/WebControls';
 import ServicesStore from '../../../stores/ServicesStore';
 import Service from '../../../models/Service';
@@ -27,6 +27,18 @@ class WebControlsScreen extends Component {
 
   autorunDisposer = null;
 
+  @action _setCanGoBack() {
+    this.canGoBack = this.webview.canGoBack();
+  }
+
+  @action _setCanGoForward() {
+    this.canGoForward = this.webview.canGoForward();
+  }
+
+  @action _setUrl(value) {
+    this.url = value;
+  }
+
   constructor(props) {
     super(props);
 
@@ -39,15 +51,15 @@ class WebControlsScreen extends Component {
     this.autorunDisposer = autorun(() => {
       if (service.isAttached) {
         this.webview = service.webview;
-        this.url = this.webview.getURL();
+        this._setUrl(this.webview.getURL());
 
         for (const event of URL_EVENTS) {
           this.webview.addEventListener(event, e => {
             if (!e.isMainFrame) return;
 
-            this.url = e.url;
-            this.canGoBack = this.webview.canGoBack();
-            this.canGoForward = this.webview.canGoForward();
+            this._setUrl(e.url);
+            this._setCanGoBack();
+            this._setCanGoForward();
           });
         }
       }
@@ -101,7 +113,7 @@ class WebControlsScreen extends Component {
     }
 
     this.webview.loadURL(url);
-    this.url = url;
+    this._setUrl(url);
   }
 
   openInBrowser() {
--- a/src/features/webControls/containers/WebControlsScreen.jsx
+++ b/src/features/webControls/containers/WebControlsScreen.jsx
@@ -27,16 +27,14 @@ class WebControlsScreen extends Component {
 
   autorunDisposer = null;
 
-  @action _setCanGoBack() {
-    this.canGoBack = this.webview.canGoBack();
-  }
-
-  @action _setCanGoForward() {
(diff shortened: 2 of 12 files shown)


vraravam:
Just finished my work - will look to download and run this code.

vraravam:
Just finished my work - will look to download and run this code.

vraravam:
I still see a lot of these warnings though
Screenshot 2022-08-01 at 10 59 23 PM

SpecialAro:
> I still see a lot of these warnings though
> Screenshot 2022-08-01 at 10 59 23 PM
> 

Haven't seen them in my logs but tomorrow will try to force them into happening so I can solve them. Maybe if I run it in macOS they will show up faster.

Thank you for taking the time to look into this @vraravam ☺️

src/features/webControls/containers/WebControlsScreen.jsx
```diff
@@ -39,15 +51,15 @@ class WebControlsScreen extends Component {
...
         for (const event of URL_EVENTS) {
           this.webview.addEventListener(event, e => {
             if (!e.isMainFrame) return;
 
-            this.url = e.url;
-            this.canGoBack = this.webview.canGoBack();
-            this.canGoForward = this.webview.canGoForward();
+            this._setUrl(e.url);
+            this._setCanGoBack();
+            this._setCanGoForward();
```
kris7t: For performance reasons, it's usually better to use coarser actions. E.g., this whole event handler calls 3 distinct actions, which cause 3 separate mobx updates. It could be turned into a single action (e.g., `this._setUrlAndHistory(e.url)`) to only cause a single update.

src/features/workspaces/store.js
```diff
@@ -114,10 +114,10 @@ export default class WorkspacesStore extends FeatureStore {
   }
 
   @action reset() {
-    this.activeWorkspace = null;
-    this.nextWorkspace = null;
+    this._setActiveWorkspace(null);
+    this._setNextWorkspace(null);
     this.workspaceBeingEdited = null;
-    this.isSwitchingWorkspace = false;
+    this._setIsSwitchingWorkspace(false);
```
kris7t, line 120: Again, we trigger a bunch of separate updates at the same time here.

src/models/Service.ts
```diff
@@ -420,26 +444,26 @@ export default class Service {
...
       debug('Did start load', this.name, event);
 
-      this.hasCrashed = false;
-      this.isLoading = true;
-      this.isLoadingPage = true;
-      this.isError = false;
+      this._setHasCrashed(false);
+      this._setIsLoading(true);
+      this._setIsLoadingPage(true);
+      this._setIsError(false);
```
kris7t: Is there a way to do these updates in batch instead?

src/models/Service.ts
```diff
@@ -420,26 +444,26 @@ export default class Service {
...
     // eslint-disable-next-line unicorn/consistent-function-scoping
     const didLoad = () => {
-      this.isLoading = false;
-      this.isLoadingPage = false;
+      this._setIsLoading(false);
+      this._setIsLoadingPage(false);
 
       if (!this.isError) {
-        this.isFirstLoad = false;
+        this._setIsFirstLoad(false);
```
kris7t: Is there a way to do these updates in batch instead (possible with moving the conditional logic into an action as well)?

src/models/Service.ts
```diff
@@ -453,16 +477,16 @@ export default class Service {
...
         event.errorCode !== -3
       ) {
-        this.isError = true;
-        this.errorMessage = event.errorDescription;
-        this.isLoading = false;
-        this.isLoadingPage = false;
+        this._setIsError(true);
+        this._setErrorMessage(event);
+        this._setIsLoading(false);
+        this._setIsLoadingPage(false);
```
kris7t: Is there a way to do these updates in batch instead?

src/features/webControls/containers/WebControlsScreen.jsx
SpecialAro: I agree :D (It's my first time doing `mobx` stuff so any help is really appreciated so I can learn how to properly do things)

Should it be something like this?
image

kris7t: yep, this looks about right. the end of each _top-level_ action should trigger a new update, so the call to `_setUrl` from `_setUrlAndHistory` won't (but calls to `_setUrl` and `_setUrlAndHistory` from event handlers will, as they should).

SpecialAro:
@kris7t I've pushed some changes. Can you see if that's what you intended with your comments?

kris7t:
yeah, looks great! :slightly_smiling_face:

vraravam:
This works great now! No more mobx warnings! Can we merge this please?

cino (approved):
Review from phone: code looks good, can't locally test it before august 11th.

SpecialAro:
> This works great now! No more mobx warnings! Can we merge this please?

I'm just gonna cleanup stuff and create smaller actions to be called by the main actions, instead of repeating code on two or three different actions.

SpecialAro:
There are still 3 warnings I can't seem to fix - if anyone wants to give it a go before the merging, please commit to this PR 😄.

The first two happen when you edit and save a Workspace and the other one when you create (or delete) a Workspace.

image

src/features/serviceProxy/index.ts
```diff
@@ -15,8 +15,12 @@ export default function init(stores: {
 }) {
   debug('Initializing `serviceProxy` feature');
 
+  const setIsEnabled = action((value: boolean) => {
```
vraravam, line 18: Is there a reason for the different style of the code? All other places its `@action <funcName>`, while here its different?

SpecialAro: I tried that and typescript didn't like it. I'm guessing `@action` only works inside a class - which is why I followed that approach.

image
image

vraravam: ok - resolving my comment. LGTM!

vraravam (approved):
LGTM!

PR #540, opened 2022-08-04
fix: dark mode recipe and app handler

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Change how dark mode is activated in recipes. The global app setting shouldn't try to set all recipes to dark and the user should manually do so.

#### Motivation and Context
While working on https://github.com/ferdium/ferdium-recipes/pull/138 I stumble across a problem when trying to activate the dark mode of a recipe while using Ferdium without Dark mode (in the global settings).

After a while at looking at the code I realized that the problem was this line I removed - it was only activating the dark theme on a recipe if `service.isDarkModeEnabled` and `app.isDarkThemeActive` where both true... which I think is unintended, given that a user using the light theme of Ferdium may want to use a dark theme recipe.

I understand that this may need further thinking in the future - maybe with something like: if the user didn't set `service.isDarkModeEnabled` and `app.isDarkThemeActive` then `service.isDarkModeEnabled` should default to `true`.

#### Checklist

- [x] `npm test` passes

#### Release Notes

fix: dark mode recipe and app handler

### DIFF
--- a/src/webview/recipe.js
+++ b/src/webview/recipe.js
@@ -334,10 +334,7 @@ class RecipeController {
       isDarkModeStyleInjected,
     };
 
-    if (
-      this.settings.service.isDarkModeEnabled &&
-      this.settings.app.isDarkThemeActive !== false
-    ) {
+    if (this.settings.service.isDarkModeEnabled) {
       debug('Enable dark mode');
 
       // Check if recipe has a custom dark mode handler


vraravam (approved):

vraravam:
imo, the 2 different variables (one global vs one per-service) can cause a lot of confusion depending on different combinations. On top of this, there's also the hidden part about whether the user chose a particular setting or whether that was defaulted. My vote is to remove the global and only keep the per-service one. Though it will mean that a user (who wants to use dark mode everywhere), will have to edit each service separately, it would make it simpler to support and will reduce bugs (imo)

SpecialAro:
> imo, the 2 different variables (one global vs one per-service) can cause a lot of confusion depending on different combinations. On top of this, there's also the hidden part about whether the user chose a particular setting or whether that was defaulted. My vote is to remove the global and only keep the per-service one. Though it will mean that a user (who wants to use dark mode everywhere), will have to edit each service separately, it would make it simpler to support and will reduce bugs (imo)

I think that the global setting on "Dark Theme" should only refer to the dark mode of our UI and not the service. So I agree with you.

Can we merge this PR?

vraravam:
Sure - lets merge

PR #698, opened 2022-10-24
chore: remove deprecated @types packages

cino:
After `npm install` I received a deprecated message for the following:

image

Based on that decided to remove these dependencies as these are deprecated.


vraravam (approved):
Haven't tested locally. But, if the app works without these, then LGTM!

PR #750, opened 2022-11-10
fix: ensure default values are working for the Toggle fields

cino:
After the changes to the Toggle component the Default values aren't displayed correctly. The `default` field does not seem to be respected since the upgrade. Decided to add the default to the value property when the value is otherwise undefined.

### DIFF
--- a/src/containers/settings/EditServiceScreen.tsx
+++ b/src/containers/settings/EditServiceScreen.tsx
@@ -193,6 +193,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isEnabled,
             DEFAULT_SERVICE_SETTINGS.isEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isEnabled,
           type: 'checkbox',
         },
         isHibernationEnabled: {
@@ -201,6 +202,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isHibernationEnabled,
             DEFAULT_SERVICE_SETTINGS.isHibernationEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isHibernationEnabled,
           type: 'checkbox',
         },
         isWakeUpEnabled: {
@@ -209,6 +211,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isWakeUpEnabled,
             DEFAULT_SERVICE_SETTINGS.isWakeUpEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isWakeUpEnabled,
           type: 'checkbox',
         },
         isNotificationEnabled: {
@@ -217,6 +220,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isNotificationEnabled,
             DEFAULT_SERVICE_SETTINGS.isNotificationEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isNotificationEnabled,
           type: 'checkbox',
         },
         isBadgeEnabled: {
@@ -225,6 +229,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isBadgeEnabled,
             DEFAULT_SERVICE_SETTINGS.isBadgeEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isBadgeEnabled,
           type: 'checkbox',
         },
         isMediaBadgeEnabled: {
@@ -233,6 +238,7 @@ class EditServiceScreen extends Component<IProps> {
             service?.isMediaBadgeEnabled,
             DEFAULT_SERVICE_SETTINGS.isMediaBadgeEnabled,
           ),
+          default: DEFAULT_SERVICE_SETTINGS.isMediaBadgeEnabled,
           type: 'checkbox',
         },
         trapLinkClicks: {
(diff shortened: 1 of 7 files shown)


cino:
Let's wait before merging, having some issues now with Toggles not sure if this change caused it.

src/containers/settings/EditServiceScreen.tsx
```diff
@@ -188,50 +188,52 @@ class EditServiceScreen extends Component<IProps> {
         },
         isEnabled: {
           label: intl.formatMessage(messages.enableService),
-          value: service?.isEnabled,
-          default: DEFAULT_SERVICE_SETTINGS.isEnabled,
+          value: service?.isEnabled || DEFAULT_SERVICE_SETTINGS.isEnabled,
```
vraravam: This will also evaluate the 2nd condition if the 1st one returns a false. But, I think you wanted to only handle the `undefined` condition, am i right?

SpecialAro:
@muhamedsalih-tw is this PR still relevant after https://github.com/ferdium/ferdium-app/pull/790 ?

ghost:
> @muhamedsalih-tw is this PR still relevant after #790 ?

@SpecialAro - yes, this PR valid and it about the ui respecting default setting.
currently in addition of new service i see only enable audio is respecting default setting.
This PR need to validated then can be used to fix the stated issue.

#790 is about a fix for a bug that default proxy data of recipe not being set properly in request cache object [Temp fix is given].

ghost:
@cino - A suggestion, I don't see point in assigning `value key` with or-ed value of (`value key` and `default key` ) and removing `default key`.

 we may need to understand if the recipe specific default setting set to `value key` properly (means does it comes properly) if not we can assign undefined to the `value key` so that it take default key directly
 
 example:
 dark theme slider repect default value in curr nightly
 enable audio respect default value in current nightly

SpecialAro:
> > @muhamedsalih-tw is this PR still relevant after #790 ?
> 
> @SpecialAro - yes, this PR valid and it about the ui respecting default setting.
> currently in addition of new service i see only enable audio is respecting default setting.
> This PR need to validated then can be used to fix the stated issue.
> 
> #790 is about a fix for a bug that default proxy data of recipe not being set properly in request cache object [Temp fix is given].
> 
> 
> 

Awesome 👍 thank you! Just wanted to double check 😁

src/containers/settings/EditServiceScreen.tsx
vraravam: @cino - could you please verify that my later commit into your PR branch - whether you think that's a better way to handle this issue?

vraravam:
> @cino - A suggestion, I don't see point in assigning `value key` with or-ed value of (`value key` and `default key` ) and removing `default key`.
> 
> we may need to understand if the recipe specific default setting set to `value key` properly (means does it comes properly) if not we can assign undefined to the `value key` so that it take default key directly
> 
> example: dark theme slider repect default value in curr nightly enable audio respect default value in current nightly

@muhamedsalih-tw - I think my latest commit is along the same lines that you suggested. Please confirm

But, at the same time, thinking of a future feature: For each field in the Settings screens, can we think of a "reset to default value" kind of an action/link? This very useful, and I have used that in VSCode. If that's something we want to explore, then this PR is not the correct implementation to enable/allow for that.

ghost:
@vraravam - I have ran the app and its running fine.
I reviewed code @cino I still strongly recommend to use default values as separate key value pair like old code that we feed all in values.

Now I see default value is removed we may need to manually do reset form or field operations in form instead of using the react-mobx-form lib.

For the above asked near future feature we can utilise the default value to reset.

For ref:
field defult value key - https://foxhound87.github.io/mobx-react-form/docs/api-reference/fields-properties.html
form default value reset - https://foxhound87.github.io/mobx-react-form/docs/actions/shared.html#clear--reset
Field level reset to default value - https://foxhound87.github.io/mobx-react-form/docs/api-reference/fields-methods.html
example - https://foxhound87.github.io/mobx-react-form/docs/fields/defining-flat-fields/unified-properties.html (under topic Define Fields with Default Value (on reset))


In Short:
what you done is correct but dont remove default values key!!

src/containers/settings/EditServiceScreen.tsx
ghost: @cino - I am suggesting to have default value and the value separately

Eg:

     isEnabled:{
          label: intl.formatMessage(messages.enableService),
          value: service?.isEnabled,
          default: DEFAULT_SERVICE_SETTINGS.isEnabled, // must be there  currently deleted
          value:  ifUndefined<boolean>(
           service?.isEnabled,
            DEFAULT_SERVICE_SETTINGS.isEnabled,
          ),, // its correct

ghost:
checked in mac m1 - works fine
image

vraravam (approved):

src/containers/settings/EditServiceScreen.tsx
```diff
@@ -188,88 +189,117 @@ class EditServiceScreen extends Component<IProps> {
...
+            service?.trapLinkClicks,
+            DEFAULT_SERVICE_SETTINGS.trapLinkClicks,
+          ),
           default: DEFAULT_SERVICE_SETTINGS.trapLinkClicks,
           type: 'checkbox',
         },
         isMuted: {
           label: intl.formatMessage(messages.enableAudio),
-          value: !service?.isMuted,
+          value: ifUndefined<boolean>(
```
Alphrag: This gives the opposite value as expected since the checkbox is about enabling the audio, but the boolean is about being muted. So this should be `! ifUndefined...` in the same way the negation was there before.

vraravam: thanks for catching that - fixed in the latest commit

Alphrag (approved):

PR #800, opened 2022-11-25
fix sqlite3 issue with electron-rebuild

balajiv113:
#### Pre-flight Checklist

#### Description of Change
Fixes the cross built binary issue

**Note:**
Not a big fan of this fix. This is done to un-block the nightly build not even loading. 
We can explore possible better options for this as well

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/electron-builder-fix-env.js
+++ b/electron-builder-fix-env.js
@@ -1,4 +1,28 @@
-// HACKTAG: Fix from https://github.com/electron-userland/electron-builder/issues/7256 to overcome issue from electron-builder for windows multi-arch builds
+const fs = require("fs");
+const path = require("path");
 exports.default = async function (context) {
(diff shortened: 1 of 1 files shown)


Alphrag:
Even though I think patching like this is very brutal and not nice, I can confirm that this fixes the names of the bindings folders for the different arch, which means they are properly included. As an example, after compiling all targets for windows on my `W10 x64` (which involved rebuilding `sqlite3` bindings for `arm64`, `x64` and `ia32` in this order), I had no trouble running the Portable-ia32, as can be seen in the following picture:
Screenshot 2022-11-26 000045

Alphrag (approved):

Alphrag:
Since this problem is quite important, I am merging this for now, but we should consider setting the `napi_build_version` variable in the process environment span by `electron-rebuild` if this is possible in the same way we remove the wrong MSVS version.

PR #803, opened 2022-11-28
Fix winget workflow for beta releases [ci skip]

Alphrag:
Fix the winget workflow that publishes the beta releases under the nightly one.


vraravam (approved):
Haven't tested this on windows machine. But, the changes look good.

PR #875, opened 2023-01-12
Use dependabot to help fixing vulnerable dependencies and improve security

abacao:
Hello developers
The use of dependabot is free for public projects/repositories.

It will help you to maintain many of your dependencies up to date resulting on a more secure application.

You need to enable some options on the repo configuration after adding this file.
(Settings -> Code security -> enable top 5 options)

At this moment, dependabot is able to create 5 PullRequests were he is suggestion you to bump some versions such as:

- eslint-plugin-jsx-a11y from 6.6.1 to 6.7.1
- eslint-plugin-import from 2.26.0 to 2.27.4
- wait-on from 6.0.1 to 7.0.1
- node-fetch from 2.6.7 to 3.3.0
- ws from 8.11.0 to 8.12.0

Also, Dependabot is alerting about 4 different vulnerabilities regarding jsonwebtoken.
The version in use is vulnerable for some different things where the highest CVSS (Common Vulnerability Scoring System) score is 7.5 and the lowest is still to be determined.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"


SpecialAro:
@vraravam what do you think about this? I don't see any disadvantage of using dependabot (I'm actually running Renovate in other cases and it works just fine)

vraravam:
If we can somehow limit it to minor / tiny version bumps, then mostly I am ok. This is because, in some cases, a version getting bumped (and which would trigger the automated tests) would not guarantee that the app continues to run. For eg, the `react-router-dom` if bumped above 6.4.2 has breaking changes. (I forget if those changes show up during the CI process or not, but definitely the app does not start up).

OTOH, I am ok to use this as a mechanism to remind the contributors that these packages have newer versions and maybe they can run those newer versions from the PRs that dependabot raises - run the app locally before approving. This can be a safe / acceptable process and use of this tool. In other words, even if we use this, we should not set it up for auto-merge. And there needs to be an agreement that all changes are built and run locally before approving the PRs.

Currently, I run `pnpm outdated` manually at random intervals and manually do the upgrades by following the above checklist steps before pushing.

vraravam (approved):

vraravam:
@all-contributors please add @victorbnl for security

marie-bnl:
Was that intended? I don't actually remember having contributed to Ferdium's security

vraravam:
yes - this PR enhances our security :)

marie-bnl:
Well it does but I didn't make this PR 😂

vraravam:
@all-contributors please add @abacao for security

PR #1150, opened 2023-05-05
Upgrade 'snapcraft' to 'v7.x'

NathanaelGandhi:
#### Pre-flight Checklist

#### Description of Change

- Updated snapcraft in the build workflow from v5.x to 7.x. 
- Updated snapcraft login to use env var.

#### Motivation and Context

- Snapcraft login secret expired. Generating one with the latest version of snapcraft requires we target  >= v7.2. Latest (at time of writing) is 7.3.x.

#### Checklist

- Unable to test GH workflows locally

#### Release Notes

Update snapcraft from v5.x to v7.x in build workflow

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -297,10 +297,9 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.FERDIUM_PUBLISH_TOKEN }}
           CSC_IDENTITY_AUTO_DISCOVERY: false
-          SNAPCRAFT_LOGIN: ${{ secrets.SNAPCRAFT_LOGIN }}
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_LOGIN }} # Note: Login no longer required https://forum.snapcraft.io/t/snapcraft-authentication-options/30473/21 
         run: |
-          sudo snap install snapcraft --channel=5.x/stable --classic
-          echo "$SNAPCRAFT_LOGIN" | snapcraft login --with -
+          sudo snap install snapcraft --channel=7.x/stable --classic
           pnpm build --publish always -c.snap.publish.repo=$SNAP_PUBLISH_REPO -c.snap.publish.channels=$SNAP_PUBLISH_CHANNEL
           snapcraft logout
         shell: bash


.github/workflows/builds.yml
```diff
@@ -299,8 +299,8 @@ jobs:
           CSC_IDENTITY_AUTO_DISCOVERY: false
           SNAPCRAFT_LOGIN: ${{ secrets.SNAPCRAFT_LOGIN }}
         run: |
-          sudo snap install snapcraft --channel=5.x/stable --classic
-          echo "$SNAPCRAFT_LOGIN" | snapcraft login --with -
+          sudo snap install snapcraft --channel=7.x/stable --classic
+          export SNAPCRAFT_STORE_CREDENTIALS=$(cat "$(SNAPCRAFT_LOGIN)")
```
Alphrag: This is not appropriate since `cat` requires a file behind, and `$( SNAPCRAFT_LOGIN )` is trying to execute the command `SNAPCRAFT_LOGIN` which does not exist whereas we only want to get the value in the variable itself.

.github/workflows/builds.yml
```diff
@@ -299,8 +299,8 @@ jobs:
           CSC_IDENTITY_AUTO_DISCOVERY: false
           SNAPCRAFT_LOGIN: ${{ secrets.SNAPCRAFT_LOGIN }}
```
Alphrag: I would simply change `SNAPCRAFT_LOGIN` by `SNAPCRAFT_STORE_CREDENTIALS` as this will store the value of the secret into the expected environment variable.

Alphrag (changes requested):

.github/workflows/builds.yml
NathanaelGandhi: Great suggestion thank you

.github/workflows/builds.yml
```diff
@@ -297,10 +297,9 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.FERDIUM_PUBLISH_TOKEN }}
           CSC_IDENTITY_AUTO_DISCOVERY: false
-          SNAPCRAFT_LOGIN: ${{ secrets.SNAPCRAFT_LOGIN }}
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_LOGIN }}
         run: |
-          sudo snap install snapcraft --channel=5.x/stable --classic
-          echo "$SNAPCRAFT_LOGIN" | snapcraft login --with -
+          sudo snap install snapcraft --channel=7.x/stable --classic
```
NathanaelGandhi, line 302: Should a comment be left here explaining that login is no longer required?

Ref: https://forum.snapcraft.io/t/snapcraft-authentication-options/30473/21 (@Alphrag thanks for finding this)

Alphrag: Yes, that would be a good idea to add this link which details what the documentation doesn't.

NathanaelGandhi: Adding the comment inside the run block didnt seem like a good idea. I've added it to the env var

Alphrag (approved):

Issue #139, opened 2022-05-11
Feature: Ability to Change Workspaces side bar size


r00tnix:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Ability to shrink or change the "Workspaces" bar.  Fully adjustable or just Words, like current to Icons would be nice to change quickly between workspaces, without opening menu which takes up a lot of space currently

### Proposed Solution

Have Icons, or ability to shrink the current Workspaces section

### Alternatives Considered

Manually open and closing the area, but does take away the convenience 

### Additional Information

<img width="419" alt="Screen Shot 2022-05-11 at 4 05 20 PM" src="https://user-images.githubusercontent.com/26657656/167947771-1462e010-72b9-4c30-a4b1-718b4d931b95.png">

Make the section in Yellow adjustable

vraravam:
I actually had raised a feature request in Ferdi for something similar. My proposal was to expand/collapse the Workspaces sidebar into icons/images that could be asociated with workspaces.

DenysMb:
What is the status of it?
I like how this feature works on Rambox (where you select an icon for each workspace and then it shows only the icon).
Is it possible to do the same on Ferdium? Let the user select an icon for each Workspace and enable an option to "show only as icon". This way the Workspace bar will be as thin as the one with the apps.

ch3vr0n5:
This is definitely something I miss from Rambox. It has some great visual enhancements like this but is much heavier on resources which is why I switched to Ferdium. 

I also think this would be a worthwhile enhancement. It would be great to be able to assign an icon/svg to a workspace and then with the "Always Show Workspace Drawer" setting, allow an additional setting to collapse/shrink the workspace drawer to icons only. If there are no icons assigned, creating a generic SVG with the first letter from the workspace, or assigning default icons would suffice.

Unfortunately my coding experiences is limited to database languages and some basic object oriented languages like .Net, python so I am not confident in my ability to contribute.

PR #1240, opened 2023-06-15
feat: change workspace names to icons

orgrimarr:
#### Pre-flight Checklist

#### Description of Change

- I reduced the size of the workspaceDrawer
- I replaced the workspaceItem text with an icon or the first letter. as a fallback
- I added in the workspace config a icon field in the form

The icon field is not persisted, I dit not manage to do it.
The icon config could be more user-friendly than a simple text input

#### Motivation and Context

This PR is a proposal regarding this issue https://github.com/ferdium/ferdium-app/issues/139 

#### Screenshots

image

image

image

#### Checklist

- [x]The changes respect the code style of the project (`pnpm prepare-code`)
- [x] `pnpm test` passes


vraravam:
@orgrimarr - is there a reason for this to be in "Draft" mode still?

src/components/ui/WorkspaceIcon.tsx
```diff
@@ -0,0 +1,71 @@
...
+    display: 'table-cell',
+    verticalAlign: 'middle',
+  },
+  icon: {
+    height: 'auto',
+    width: `${theme.serviceIcon.width}px`,
+    maxHeight: `${theme.serviceIcon.width}px`,
+  },
+  iconLetterContainer: {
+    border: '1px solid white',
```
vraravam: Some users might prefer the current style of showing the full name of the workspace rather than just a single character. Based on their choice, this also has to be dynamically calculated at runtime.

src/features/workspaces/components/WorkspaceDrawer.tsx
```diff
@@ -88,9 +87,9 @@ const styles = theme => ({
   },
 });
 
-interface IProps extends WithStylesProps<typeof styles>, WrappedComponentProps {
-  getServicesForWorkspace: (workspace: Workspace | null) => string[];
-}
+interface IProps
+  extends WithStylesProps<typeof styles>,
```
vraravam: The `.editorconfig` setting should allow longer lines. Have you run the formatter and does it still separate this declaration into multiple lines?

src/themes/default/index.ts
```diff
@@ -205,8 +205,8 @@ export default (brandPrimary: string) => {
         listItems: cloneDeep(services.listItems),
       },
       drawer: {
-        width: 300,
-        padding: 20,
+        width: 68,
```
vraravam: Some users might prefer the current style of showing the full name of the workspace rather than just a single character. Based on their choice, this also has to be dynamically calculated at runtime.

vraravam (changes requested):
Great initiative! Please continue with this, I will close my draft PR

vraravam:
1. Please create your branch from the latest version of the upstream's main/master (in this case develop) branch
2. When contributing to OSS, please check if other PRs are open for the same feature by someone else. Maybe if one exists, then you can collaborate in the same PR as opposed to creating a competing implementation.

orgrimarr:
> * Please create your branch from the latest version of the upstream's main/master (in this case develop) branch
> * When contributing to OSS, please check if other PRs are open for the same feature by someone else. Maybe if one exists, then you can collaborate in the same PR as opposed to creating a competing implementation.

Hi,

Thanks for your fast review, 😀

The goal was not to create a competing implementation, I just thought is was an easy way to share code.
That why this PR is still in draft mode.

Sorry for that 😕

orgrimarr:
Some questions 😀 :

In order allow both visual, should I add an option in the sidebar setting like "Use workspace drawer icon style" ?
In the text mode, should we display an icon ? or only if configured ? or not ?, if yes then the setting might be "Hide workspace drawer text"

How can I persist theses settings ? (icon path, drawer visual) ?
I din't understand what should I do in the code 

I'm only using Ferdium without an account. I wonder how to sync the icon setting. Because if we use a local path, the image might not exist in another host, and the OS path system might be different. Should we store the image content ? (We can ask for a specific image format and size like Rambox to reduce the size to a acceptable one)

vraravam:
> In order allow both visual, should I add an option in the sidebar setting like "Use workspace drawer icon style" ? In the text mode, should we display an icon ? or only if configured ? or not ?, if yes then the setting might be "Hide workspace drawer text"

Add this user preference in the workspace preference screen. I would also move the icon text box outside of the "create" action, and into the screen where the services are associated with a workspace.

> How can I persist theses settings ? (icon path, drawer visual) ? I din't understand what should I do in the code
> I'm only using Ferdium without an account. I wonder how to sync the icon setting. Because if we use a local path, the image might not exist in another host, and the OS path system might be different. Should we store the image content ? (We can ask for a specific image format and size like Rambox to reduce the size to a acceptable one)

The service and similarly the workspace preferences are ultimately stored in a sqlite db - whether its using the "accountless" option (which is basically running a local server), or the hosted server solution. The internal servers' code changes are in the same repo, but the hosted server is in a sibling repo called ferdium-server. I can help review/point you to the code file where the DML needs to be changed to add a new column in that corresponding table.

orgrimarr:
Thanks :) 

I will add a button in the user preference  😀

>  I would also move the icon text box outside of the "create" action, and into the screen where the services are associated with a workspace.

What do you mean by 'move the icon text box outside of the "create" action' ?
Should I remove the icon input in the create view like this ? 
So the icon can only be edited where the services are associated with a workspace

image

> I can help review/point you to the code file where the DML needs to be changed to add a new column in that corresponding table.

Yes please, can you help me on that point ? Which file should I modify to add
- The new icon property on the Workspace data
- The new user setting property to toggle the icon view

src/features/workspaces/components/WorkspaceDrawer.tsx
orgrimarr: Yes it is my formatter that have splited into multiple lines. 
Maybe we can add an `printWidth` setting in `.prettierrc.js`so everyone have the same setting
https://prettier.io/docs/en/options.html#print-width

orgrimarr:
It should work as discussed. 

Just missing the workspace iconUrl persistance. 
The user setting seams to be automagically persisted 😀

src/components/layout/AppLayout.tsx
```diff
@@ -53,12 +53,21 @@ const styles = (theme: { workspaces: { drawer: { width: any } } }) => ({
     // width: `calc(100% + ${theme.workspaces.drawer.width}px)`,
     width: '100%',
     transition,
+  },
+  appContentTransformTextWorkspace: {
     transform() {
       return workspaceStore.isWorkspaceDrawerOpen
```
vraravam, line 59: Would it be better to move the ternary to be evaluated inside the `translateX()` like so?
```
return 'translateX(`workspaceStore.isWorkspaceDrawerOpen ? 0 : -75px`)');
```
(Syntax might be incorrect since I'm not typing inside an editor)

src/components/layout/AppLayout.tsx
```diff
@@ -53,12 +53,21 @@ const styles = (theme: { workspaces: { drawer: { width: any } } }) => ({
...
         ? 'translateX(0)'
         : `translateX(-${theme.workspaces.drawer.width}px)`;
     },
   },
+  appContentTransformIconWorkspace: {
+    transform() {
+      return workspaceStore.isWorkspaceDrawerOpen
```
vraravam, line 66: same here.

src/containers/settings/EditSettingsScreen.tsx
```diff
@@ -259,6 +259,10 @@ const messages = defineMessages({
     id: 'settings.app.form.alwaysShowWorkspaces',
     defaultMessage: 'Always show workspace drawer',
   },
+  useWorkspaceDrawerIconStyle: {
+    id: 'settings.app.form.useWorkspaceDrawerIconStyle',
+    defaultMessage: 'Use workspace drawer icon style',
```
vraravam, line 268: So, I was suggesting to have the user choose whether they want the name or the first character of the workspace name AND/OR the icon. Could the feature be imagined like that?

src/components/layout/AppLayout.tsx
orgrimarr: Sure, we can do that

src/components/layout/AppLayout.tsx
orgrimarr: I'm not sure that it is more readable. 
If it's ok for you, I can push that change

```javascript
  appContentTransformTextWorkspace: {
    transform() {
      return `translateX(${
        workspaceStore.isWorkspaceDrawerOpen
          ? '0'
          : `-${theme.workspaces.drawer.width}px`
      })`;
    },
  },
  appContentTransformIconWorkspace: {
    transform() {
      return `translateX(${
        workspaceStore.isWorkspaceDrawerOpen ? '0' : `-75px`
      })`;
    },
  },
```

src/containers/settings/EditSettingsScreen.tsx
orgrimarr: If I understand correctly. 

The letter is currently a fallback for workspaces that don't have an icon.

The setting should be a select with theses choices :

- text  (current display)
- (letter OR icon) AND text
- letter AND icon (if icon configured)
- letter OR icon  (current PR)

It means, 3 different workspace drawer sizes 

- the current (300px by default)
- icon OR letter only (35 - 75px)
- letter + icon (let's say 75 - 150px)

orgrimarr: Should the workspace drawer bar be the same size of the service bar ? 
I means use the serviceRibbonWidth parameter for both workspace and service bar (only when displaying icon)

orgrimarr: @vraravam Can you confirm that I have understood correctly ?
If yes, I will start the dev this week   😀

vraravam: we don't need to complicate so much. only 2 options that i was suggesting:
1. workspace name (current)
2. icon or first character of workspace name (if the user wants to use a smaller-width option for the workspace drawer)

orgrimarr: Okey perfect.😀

The current state of the code pushed for this PR match what you are describing. 
I've added an option to choose between these two modes. 

image



I just need to permanently store the icon field. 
Can you point me which files should I change to make this new workspace field persistent ?
I may need to do it for the local and distant server if I understand correctly.

SpecialAro: To permanently save this field you would probably need to sync with the server (internal and external) if I'm correct.

I'm not recalling if the server is saving workspace data as a plain JSON or not (only by checking the endpoint we can be sure of that). If that is not the case, you would need to touch both the server and the internal-server... but with the ongoing Adonis update this could be a hassle :/ I can help on that tho.

SpecialAro: Also, can we make this option the default one? I like this new look (even though I'm not using workspaces).

Great work!!

vraravam: yes, you will need to sync to the internal and external servers. The db schema is the same across both - and there's a table called `workspaces` to store this info.
Screenshot 2023-09-05 at 8 18 24 AM

As seen in the above pic (taken from my internal server), you can either add a new column for the `useIcon` or you can add it into the `data` column which holds JSONB datatype. (I personally prefer option 1.

SpecialAro: > yes, you will need to sync to the internal and external servers. The db schema is the same across both - and there's a table called `workspaces` to store this info. Screenshot 2023-09-05 at 8 18 24 AM
> 
> As seen in the above pic (taken from my internal server), you can either add a new column for the `useIcon` or you can add it into the `data` column which holds JSONB datatype. (I personally prefer option 1.

Even though I would prefer option 1 (just like you) I would advise into adding it to "data" column (but first lets be sure what type of data is being retrieved there) so that it doesn't make the migration to the new Adonis's even more complicated (as we would to update both the code in the current ferdium-server, internal server and on the current adonis update PR)

src/i18n/locales/fr.json
SpecialAro, line 1: I wouldn't worry to much on the i18n translations as we have Crowdin for those - after merging this PR we can easily translate things over there.

SpecialAro:
Hi @orgrimarr I've been testing your code and it seems that it breaks functionality if you use an horizontal bar and not a vertical one. Can you please fix that?

Also, I've take a look at how we update workspaces and, indeed, we are going to need to touch both the internal server and Ferdium server after all (no escape from that). So there is that... I can help you with that tho!

orgrimarr:
Thanks you both for your comments !
I will fix this bug and will modify the server 😃

src/containers/settings/EditSettingsScreen.tsx
orgrimarr: As suggested, i added the iconUrl field in `data`

WIP related PR on the server : https://github.com/ferdium/ferdium-server/pull/102

gogosese:
Ferdium is really powerful, and I am still learning it. I am still looking for solutions when I encounter problems. When I was chatting, Google Translate couldn't be used normally.

vraravam:
@orgrimarr - is this PR mergeable at least for the internal server bits which live in this codebase itself? Can you please help to take this to completion?

orgrimarr:
Hi,
Yes, I think it is. 
I will do a quick check this week

orgrimarr:
@vraravam I check the code quickly, looks good for me.
Behaviors sounds good, test (different size of sidebar in combination with horizontal / vertical style.

I notice that we can also use icon url, wich can be easier to configure than local file.

Should be mergeable.

SpecialAro:
@vraravam I think we should also build the logic for the ferdium server part, otherwise this feature won't work (and probably will break) for users using Ferdium server

vraravam:
Thanks for the confirmation @orgrimarr 

@SpecialAro - we currently support 4 server backends: 
1. Franz
2. Ferdi
3. Ferdium (hosted)
4. no server

This change is going to break for the first 2 anyways. The 3rd one, while under our control, still has a ways to go. With this being the case, how shall we proceed?

vraravam:
> I notice that we can also use icon url

yes - I would prefer this option as opposed to uploading an icon. For the rare cases where folks create and use custom icons, they can self-host in some website, and provide that url. In most cases though, icons are usually available on the net and so a url is easier for us to keep track of/maintain that a binary file. This will also reduce the burden on syncing between the different server backends.

@orgrimarr - could you please make such a change within this same PR?

orgrimarr:
> > I notice that we can also use icon url
> 
> yes - I would prefer this option as opposed to uploading an icon. For the rare cases where folks create and use custom icons, they can self-host in some website, and provide that url. In most cases though, icons are usually available on the net and so a url is easier for us to keep track of/maintain that a binary file. This will also reduce the burden on syncing between the different server backends.
> 
> @orgrimarr - could you please make such a change within this same PR?

@vraravam I think no changes are required.

Currently, we can only configure strings https://github.com/ferdium/ferdium-app/pull/1240/files#diff-e8933cda65f8319a0a3982de972254fd0ef8f96dac31e0530660e5608682ef3bR10
- url
- local file path

This property is used to configure the image src https://github.com/ferdium/ferdium-app/pull/1240/files#diff-46ab6517cb2faf44c8255f0b1a0955a5b534eaa2786fc1f1fbbe97e85d16942cR51

SpecialAro:
> Thanks for the confirmation @orgrimarr
> 
> @SpecialAro - we currently support 4 server backends:
> 
> 1. Franz
> 2. Ferdi
> 3. Ferdium (hosted)
> 4. no server
> 
> This change is going to break for the first 2 anyways. The 3rd one, while under our control, still has a ways to go. With this being the case, how shall we proceed?


The first two (I think the 2nd one we are not supporting anymore) are going to break regardless - and users are aware of this (a message pops-up when they change to the first two. But, given that we are referring to the same project (Ferdium) I have the opinion that we should make this feature work across our entire ecosystem (account and accountless).

In fact, if this works for the internal server (which still lacks the adonis update) then it should work for ferdium server, given that the implementation is rather similar (almost copy paste on the code).

orgrimarr:
@vraravam @SpecialAro  Regarding the url persistance within the Workspace table.

Now that the adonis migration is done, should i keep saving the url in the data column ? or create a new one ?

I can try to modify ferdium-server this week

vraravam:
> Now that the adonis migration is done, should i keep saving the url in the data column ? or create a new one ?

Keeping the same schema will hopefully keep us backward compatible with all the currently-supported servers (franz, ferdi, ferdium hosted, no-server). So I would not change that db schema at this point.

vraravam:
> Currently, we can only configure strings https://github.com/ferdium/ferdium-app/pull/1240/files#diff-e8933cda65f8319a0a3982de972254fd0ef8f96dac31e0530660e5608682ef3bR10

yes, the datatype in the codebase is a string - but, at the UI level, I think its expects a file to be dragged and dropped. How will the end user experience be if they want or are forced to use a url?

orgrimarr:
> > Currently, we can only configure strings https://github.com/ferdium/ferdium-app/pull/1240/files#diff-e8933cda65f8319a0a3982de972254fd0ef8f96dac31e0530660e5608682ef3bR10
> 
> yes, the datatype in the codebase is a string - but, at the UI level, I think its expects a file to be dragged and dropped. How will the end user experience be if they want or are forced to use a url?

At the UI level it look like this 
image

Maybe i can iprove the placeholder with something like "Icon url" or "Icon url or local path"

vraravam:
yes, a more explicit text would be very much appreciated. Sorry, I was confusing with the service icon, then realized you are referring to the workspace icon.

vraravam:
@orgrimarr - could you please resolve the conflicts?

@orgrimarr / @SpecialAro : is this PR ready to be merged (once the above is done)?

vraravam:
Closing due to inactivity. Its been 1 year since the PR was opened, and no response from the OP.

orgrimarr:
Hi @vraravam this PR is ready for merge. (Just need to rebase and fix conflicts as you asked)

There were some discussion on the server / database side regarding backward compatibility : https://github.com/ferdium/ferdium-server/pull/102

vraravam:
> Hi @vraravam this PR is ready for merge. (Just need to rebase and fix conflicts as you asked)
> 
> There were some discussion on the server / database side regarding backward compatibility : [ferdium/ferdium-server#102](https://github.com/ferdium/ferdium-server/pull/102)

please rebase, and then we can reopen the PR.

PR #1258, opened 2023-06-30
Remove the code signing of builds on windows

Alphrag:
#### Pre-flight Checklist

#### Description of Change

Remove the code signing step during the build on Windows.

#### Motivation and Context

The certificate we had to code sign our builds for Windows expired this week, which now prevents the build from completing on Windows. 

#### Checklist

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -383,8 +383,6 @@ jobs:
         if: ${{ env.GIT_BRANCH_NAME == 'nightly' || env.GIT_BRANCH_NAME == 'release' }}
         env:
           GH_TOKEN: ${{ secrets.FERDIUM_PUBLISH_TOKEN }}
-          CSC_IDENTITY_AUTO_DISCOVERY: true
-          WIN_CSC_LINK: ${{ secrets.WIN_CSC_LINK }}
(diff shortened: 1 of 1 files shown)


vraravam (approved):

fsacer:
isn't this kind of a security issue? I've always expected everything to be signed.

vraravam:
Yes its kind of a security issue; or more specifically, the individual has to trust and verify that the checksums are correct. Unfortunately, we don't have contributors for the $$ needed to get the signing certs.

Alphrag:
> isn't this kind of a security issue? I've always expected everything to be signed.

The changes made last June by the certificate authorities forcing the use of hardware keys to hold code signing certificates mean that they have become very expensive to get and need a manual intervention by its owner at each build phase. For these two reasons, we have not been able to keep with signing the releases for Windows. So yes, it is a bit of a security downgrade, and using a self-signing certificate would only make things worse.

I have been looking into the [Sigstore](sigstore.dev) project, to see whether we could use it in order to sign our releases and to make them verifiable, but I haven't been able to make a full proof of concept to work with `electron-builder` yet. But if you know of other means, feel free to let us know (if so, it would be better to start a [discussion](https://github.com/ferdium/ferdium-app/discussions) and tag me there).

PR #1282, opened 2023-07-24
chore: silence deprecated version and unmet peer dependency warnings

mcmxcdev:
#### Pre-flight Checklist

#### Description of Change
- add logged deprecated dependencies to `pnpm.allowedDeprecatedVersions`
- configure unmet peer dependencies in `pnpm.peerDependencyRules.allowedVersions`
- update used flags in `.npmrc`

#### Motivation and Context
Warning logs for deprecated dependencies are popping up on every pnpm install and can be distracting for developers. Instead, now they are silenced and whenever specific dependencies are upgraded, it can be checked if entries from `pnpm.allowedDeprecatedVersions` can be removed. Pretty much the same is true for dependencies in `pnpm.peerDependencyRules.allowedVersions`

#### Screenshots
Before, unmet peer dependency warnings were ignored because they were throwing a warning instead of error only
Screenshot from 2023-07-24 12-56-26

After, no more peer dependency warnings and they would lead to exit code 1

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,5 +1,4 @@
 save-exact = true
 engine-strict = true
 unsafe-perm = true
-legacy-peer-deps = true
-frozen-lockfile = false
+strict-peer-dependencies=true
--- a/package.json
+++ b/package.json
@@ -185,6 +185,30 @@
   "pnpm": {
     "overrides": {
       "usocket": "_EXCLUDED_"
+    },
+    "peerDependencyRules": {
+      "allowedVersions": {
+        "esbuild-sass-plugin>esbuild": "^0.16.7",
+        "react-electron-web-view>react": "^18.0.0",
+        "react-electron-web-view>react-dom": "^18.0.0",
+        "react-loader>react": "^18.0.0",
+        "react-loader>react-dom": "^18.0.0",
+        "react-sortable-hoc>react": "^18.0.0",
+        "react-sortable-hoc>react-dom": "^18.0.0"
+      }
+    },
+    "allowedDeprecatedVersions": {
+      "kleur": "2.0.2",
+      "set-value": "1.0.0",
+      "request": "2.88.2",
+      "har-validator": "5.1.5",
+      "uuid": "3.4.0",
+      "@sinonjs/fake-timers": "10.2.0",
+      "@npmcli/move-file": "1.1.2 || 2.0.1",
+      "source-map-resolve": "0.5.3",
+      "source-map-url": "0.4.1",
+      "resolve-url": "0.2.1",
+      "urix": "0.1.0"
     }
   },
   "overrides": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5378,7 +5378,7 @@ packages:
   /esbuild-sass-plugin@2.10.0(esbuild@0.16.17):
     resolution: {integrity: sha512-STv849QGT8g77RRFmroSt4VBVKjv+dypKcO4aWz8IP4G5JbRH0KC0+B8ODuzlUNu9R5MbkGcev/62RDP/JcZ2Q==}
     peerDependencies:
-      esbuild: ^0.18.0
+      esbuild: ^0.18.0 || ^0.16.7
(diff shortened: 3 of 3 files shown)


.npmrc
```diff
@@ -1,5 +1,4 @@
 save-exact = true
 engine-strict = true
 unsafe-perm = true
-legacy-peer-deps = true
-frozen-lockfile = false
```
mcmxcdev, line 5: `false` is already the default anyways, and for CI environment it makes sense to be true instead: https://pnpm.io/cli/install#--frozen-lockfile

.npmrc
```diff
@@ -1,5 +1,4 @@
...
+strict-peer-dependencies=true
```
mcmxcdev, line 4: Unmet peer dependency warnings thrown by pnpm were ignored completely. Now with `strict-peer-dependencies` enabled, it will be handled stricter: https://pnpm.io/npmrc#strict-peer-dependencies

package.json
```diff
@@ -181,6 +181,30 @@
   "pnpm": {
     "overrides": {
       "usocket": "_EXCLUDED_"
+    },
+    "peerDependencyRules": {
+      "allowedVersions": {
+        "esbuild-sass-plugin>esbuild": "^0.16.7",
```
mcmxcdev, line 191: `esbuild-sass-plugin´ requires v0.18.0 but the upgrade path contains quite a lot of breaking changes that have to be handled.

package.json
```diff
@@ -181,6 +181,30 @@
...
+        "react-electron-web-view>react": "^18.0.0",
+        "react-electron-web-view>react-dom": "^18.0.0",
+        "react-loader>react": "^18.0.0",
+        "react-loader>react-dom": "^18.0.0",
+        "react-sortable-hoc>react": "^18.0.0",
+        "react-sortable-hoc>react-dom": "^18.0.0"
```
mcmxcdev, line 197: All these dependencies are pretty much deprecated/unmaintained (don't support React 18) and should be switched out with alternatives.

package.json
```diff
@@ -181,6 +181,30 @@
...
+      "set-value": "1.0.0",
+      "request": "2.88.2",
+      "har-validator": "5.1.5",
+      "uuid": "3.4.0",
+      "@sinonjs/fake-timers": "10.2.0",
+      "@npmcli/move-file": "1.1.2 || 2.0.1",
+      "source-map-resolve": "0.5.3",
+      "source-map-url": "0.4.1",
+      "resolve-url": "0.2.1",
+      "urix": "0.1.0"
```
mcmxcdev, line 211: These warnings are now silenced, and they should be worked down one by one ideally.

It is possible to trace back which dependencies need to be upgraded by running e.g. `pnpm why uuid`. In that case, `@adonisjs/framework` is the culprit.

package.json
vraravam: please add TODO comments in the code - once this PR is closed, we will lose track of such info

package.json
vraravam: please add TODO comments in the code - once this PR is closed, we will lose track of such info

package.json
vraravam: please add TODO comments in the code - once this PR is closed, we will lose track of such info

vraravam (changes requested):
please add TODO comments in the code as per the previous comments

package.json
mcmxcdev: `package.json` doesn't allow comments unfortunately, where would you like me to add this?

package.json
mcmxcdev: `package.json` doesn't allow comments unfortunately, where would you like me to add this?

vraravam: shall we create issues so that this is not lost?

mcmxcdev: Yeah, that was my plan, I will create an issue for it, maybe later in the day.

UPDATE: https://github.com/ferdium/ferdium-app/issues/1284

vraravam (approved):

PR #1283, opened 2023-07-25
style: reformat all files properly

mcmxcdev:
#### Pre-flight Checklist

#### Description of Change
- remove `.json` from `.eslintignore`
- add type intellisense to `.prettierrc.js`
- install missing `@jest/types` which is used in `jest.config.js` for type intellisense
- split up `typecheck` and `lint` into two separate commands
- remove `--require-pragma` flag from `reformat-files` which leads to unformatted files
- install suggested `@types/prop-types` and `@types/react-loader` dependencies
- add `@ts-expect-error` to `Loader` usages
- reformat all scss files automatically with `reformat-files` command

#### Motivation and Context
The project needs to reach a state where all files are properly formatted and we profit from type intellisense where possible.

#### Checklist

- [x] `pnpm test` passes


src/components/ui/InfoBar.tsx
```diff
@@ -72,6 +72,7 @@ class InfoBar extends Component<IProps> {
                 lines={10}
                 scale={0.3}
                 color="#FFF"
+                // @ts-expect-error Property 'component' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<ReactLoader> & Readonly<LoaderProps>
```
mcmxcdev, line 75: These pop up now because I installed `@types/react-loader`. I wouldn't bother with them at all, since `react-loader` is a deprecated dependency that we want to replace.

package.json
```diff
@@ -30,15 +30,16 @@
...
-    "lint": "tsc --noEmit && eslint \"{src,test,scripts}/**/*.{js,jsx,mjs,ts,tsx}\"",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint \"{src,test,scripts}/**/*.{js,jsx,mjs,ts,tsx}\"",
     "lint:fix": "eslint --fix \"{src,test,scripts}/**/*.{js,jsx,mjs,ts,tsx}\"",
     "extract": "formatjs extract \"src/**/*.{js,jsx,mjs,ts,tsx}\" --ignore=\"**/*.d.ts\" --out-file temp.json --flatten --id-interpolation-pattern '[sha512:contenthash:base64:6]' --preserve-whitespace",
     "compile": "formatjs compile \"temp.json\" --out-file src/i18n/locales/en-US.json",
     "manage-translations": "pnpm extract && pnpm compile && rimraf temp.json",
     "build": "preval-build-info-cli && node esbuild.mjs && electron-builder",
     "contributors": "all-contributors",
-    "reformat-files": "prettier --ignore-path .eslintignore --write --require-pragma \"**/*.{js,jsx,mjs,ts,tsx,scss}\"",
```
mcmxcdev, line 40: I dont know why `--require-pragma` flag was ever put into the project, but it shouldn't be here because "prettier will restrict itself to only format files that contain a special comment": https://prettier.io/docs/en/options.html#require-pragma

package.json
```diff
@@ -30,15 +30,16 @@
     "dev": "node esbuild.mjs --watch",
     "test": "jest",
     "test:watch": "jest --watch",
-    "lint": "tsc --noEmit && eslint \"{src,test,scripts}/**/*.{js,jsx,mjs,ts,tsx}\"",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint \"{src,test,scripts}/**/*.{js,jsx,mjs,ts,tsx}\"",
```
mcmxcdev, line 34: These two should be separate, since a project can be conformed to TS but not to lint rules. Always running type checks when linting is extra effort and prolongs the command execution too.

src/components/ui/InfoBar.tsx
vraravam: so, why do we need to introduce this `react-loader` at this point if we know that its to be replaced? Wasn't the codebase working without it?

mcmxcdev: In general, it makes sense to install all `@types` dependencies which accompany the dependencies which are written in JS. Although we want to replace `react-loader`, it is unclear when this will happen so introducing it here makes sure that in case new usages of `react-loader` are used, they have to conform to the type definitions.

vraravam (approved):

PR #1292, opened 2023-07-27
chore: upgrade mobx-react-form to latest

mcmxcdev:
#### Pre-flight Checklist

#### Description of Change
- upgrade `mobx-react-form` from v3 to v6
- remove stub `mobx-react-form.d.ts` since the dependency is written in TS now
- remove `as any` cast for `new Form` invocation in `invite.tsx`
- reuse `Error` component in `Radio.tsx` to get validation error message in red
- use `FieldInterface` type for `field` property instead of `typeof Field` in `Radio.tsx`

#### Motivation and Context
The dependency was a couple of major versions behind, luckily the upgrade path was quite easy. The new major versions all seem to have to do with the library adopting TypeScript:
https://github.com/foxhound87/mobx-react-form/releases/tag/v4.0.0
https://github.com/foxhound87/mobx-react-form/releases/tag/v5.0.0
https://github.com/foxhound87/mobx-react-form/releases/tag/v6.0.0

#### Screenshots

`Form` and `Radio` components which are using `mobx-react-form` are still working as expected:

[Screencast from 2023-07-26 19-56-59.webm](https://github.com/ferdium/ferdium-app/assets/16797721/6aaf3a9d-fdf0-466f-a550-01759610c50f)

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/src/components/ui/Radio.tsx
+++ b/src/components/ui/Radio.tsx
@@ -1,10 +1,11 @@
 import { Component } from 'react';
 import { observer } from 'mobx-react';
-import { Field } from 'mobx-react-form';
 import classnames from 'classnames';
+import FieldInterface from 'mobx-react-form/lib/models/FieldInterface';
+import Error from './error';
 
 type Props = {
-  field: typeof Field;
+  field: FieldInterface;
   className: string;
   focus: boolean;
   showLabel: boolean;
@@ -47,7 +48,8 @@ class Radio extends Component<Props> {
           </label>
         )}
         <div className="franz-form__radio-wrapper">
-          {field.options.map(type => (
+          {/* @ts-expect-error Property 'map' does not exist on type 'OptionsModel'. */}
+          {field.options?.map(type => (
             <label
               key={type.value}
               htmlFor={`${field.id}-${type.value}`}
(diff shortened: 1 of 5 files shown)


src/components/ui/Radio.tsx
```diff
@@ -47,7 +48,8 @@ class Radio extends Component<Props> {
           </label>
         )}
         <div className="franz-form__radio-wrapper">
-          {field.options.map(type => (
+          {/* @ts-expect-error Property 'map' does not exist on type 'OptionsModel'. */}
+          {field.options?.map(type => (
```
mcmxcdev, line 52: I think an arbitrary naming of `options` was chosen on Franz/Ferdi side a long time ago to be passed in as possible radio options.

Unfortunately, it clashes with the naming of some `mobx-react-form` specific type, but we can ignore it.

vraravam: ok - as long as its working for now, its fine.
But, do you think we should put a `NOTE` or some other comment with this text for future reference?

vraravam (changes requested):
please see earlier comment

src/components/ui/Radio.tsx
mcmxcdev: This LOC will always lead back to this PR, so that's also one way to read up on why the TS error is ignored.

Issue #1105, opened 2023-04-10
No notifications 


vncloudsco:
### Your issue

we don't receive any notifications from apps,  all messages on my apps are not receiving any notifications, the number of incoming messages is displayed next to the icon but there is no notification of new messages how can i fix this problem?

marie-bnl:
What service are you expecting getting notifications from? Ferdium by itself does not send notifications. Thus you can only get notifications from services that send them, i.e. services you get notifications from on the web browser. If a website does not send web notifications, Ferdium can’t do it for it. This is because it has no system to read the messages’ sender information and content.

collindutter:
@victorbnl I'm not receiving notifications on Gmail. A notification badge appears on the Ferdium icon, but I receive no alerts. This is on macOS 13.3, Ferdium 6.2.6.

marie-bnl:
Do you receive notifications from GMail on your web browser?

gschafra:
Same here for WhatsApp, see #1182. After fixing the not working message count in the badge the missing notifications problem popped up, which I've been trying to solve. And then, while looking into the API doc and the code, I realized that Ferdium does not send notifications but the service does or should do....

akatarmo:
> Ferdium by itself does not send notifications. Thus you can only get notifications from services that send them, i.e. services you get notifications from on the web browse

Since Ferdi died I have been using Ferdium. And I have **always** received WhatsApp notifications. But it's been about a month since the notification banner in the statusbar on the right is not displayed.

And of course Whatsapp sends them because I am now using **Rambox** until the problem is fixed in Ferdium and in Rambox I receive notifications as expected.

marie-bnl:
I was able to reproduce the issue, WhatsApp does send notifications on Firefox and not on Ferdium. I’ll try to investigate it.

marie-bnl:
So, here is the state of affairs. Some services do send notifications as expected, I receive notifications from ProtonMail. Some don’t send notifications but it seems to be the website’s issue, on GMail I don’t receive notifications, but neither do I on Firefox. Last, interestingly, WhatsApp does send notifications on Firefox and not on Ferdium. I wonder if it happens with other services as well. In the case of WhatsApp I noticed updates don’t work (click update and refresh the page, the update button will still be there, which doesn’t happen on Firefox), and I wonder if it could be the cause of the issue.

PingWeb:
[...]

PR #1357, opened 2023-09-12
workaround: notifications not displaying for WhatsApp

agreenbhm:
#### Pre-flight Checklist

#### Description of Change
Added try/catch for displaying notifications, with notification options object setting onClick to null in the event of error.  Fixes lack of notifications for WhatsApp (and possibly others) due to failure to clone the options object while displaying notifications (related to onClick function).  This is a workaround until the full problem is resolved, as the change allows notifications to be displayed but when clicking them you are not taken to the proper screen (due to null'ing the onClick function).  This is much better than not receiving any notifications while waiting for a full fix, IMO.

#### Motivation and Context
Workaround/Fix for https://github.com/ferdium/ferdium-app/issues/1105 - WhatsApp notifications are not being displayed.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

Workaround for notifications not displaying for some apps (WhatsApp + more).

### DIFF
--- a/src/webview/notifications.ts
+++ b/src/webview/notifications.ts
@@ -46,7 +46,7 @@ export const notificationsClassDefinition = `(() => {
           });
       } catch(error) {
 	        this.options.onClick = null;
-          window.ferdium.displayNotification(this.title, this.options)
+          window.ferdium.displayNotification(title, options)
             .then(() => {
               if (typeof (this.onClick) === 'function') {
                 this.onClick();
--- a/src/webview/notifications.ts
+++ b/src/webview/notifications.ts
@@ -37,12 +37,22 @@ export const notificationsClassDefinition = `(() => {
     constructor(title = '', options = {}) {
       this.title = title;
       this.options = options;
-      window.ferdium.displayNotification(title, options)
-        .then(() => {
-          if (typeof (this.onClick) === 'function') {
-            this.onClick();
-          }
-        });
+      try{
+        window.ferdium.displayNotification(title, options)
+          .then(() => {
+            if (typeof (this.onClick) === 'function') {
+              this.onClick();
+            }
+          });
+      } catch(error) {
+	        this.options.onClick = null;
+          window.ferdium.displayNotification(this.title, this.options)
+            .then(() => {
+              if (typeof (this.onClick) === 'function') {
+                this.onClick();
+              }
+            });      
+      }
     }
 
     static requestPermission(cb = null) {


src/webview/notifications.ts
```diff
@@ -37,12 +37,22 @@ export const notificationsClassDefinition = `(() => {
...
+          });
+      } catch(error) {
+	        this.options.onClick = null;
+          window.ferdium.displayNotification(title, options)
+            .then(() => {
+              if (typeof (this.onClick) === 'function') {
+                this.onClick();
+              }
+            });      
+      }
```
mcmxcdev, line 55: I think you got confused a bit with the try catch and then/catch blocks, With your code, the same method would be called in both try and catch. 

What do you think about this:

```suggestion
try {
  window.ferdium.displayNotification(title, options);

  if (typeof this.onClick === 'function') {
    this.onClick();
  }
} catch {
  this.options.onClick = null;
}
```

mcmxcdev (commented):
Not sure if we want to merge this including the "null'ing the onClick function" workaround since it doesn't redirect to the correct screen as you mentioned. 

What do you think @vraravam ?

src/webview/notifications.ts
agreenbhm: I understand why it is seen as a bit repetitive, however I think it may be necessary.  IIRC, 'this.onClick' will still return as being typeof function even in the failure scenario, but when the displayNotifications function is called the "options" object seems to fail to be copied for some reason (something seemingly to do with the onClick function).  My point is, the error occurs further down the line so we can't just assume that it showing as typeof function is enough for it to succeed.

agreenbhm: Just checked, and as I said, it seems that typeof returns the correct value even if it will then go on to fail.  See screenshot below.  First debug console command was run at the breakpoint, and then when execution is resumed the error is thrown.

image

mcmxcdev: Thanks for the thorough investigation! It seems it has to be written that way, so fine with continuing. 

Still would like to defer to @vraravam

vraravam: I am not enough of a js developer to comment on this nuance. I will defer to you and other core committers for the decision.

agreenbhm: I'm no JS expert either. That said, this seems to mostly fix a significant issue even if it's a bit of a dirty hack, so I hope you will approve it.

mcmxcdev (approved):

src/webview/notifications.ts
SpecialAro, line 55: Any reason we are having a try catch block instead of (async function).then().catch() ?

I agree everything is a bit of an hack and we should probably make some sense of it in the future.

EDIT: also, just noticed we are repeating code in the catch section. Maybe we should move that logic to its own function, no?

PR #1507, opened 2024-01-01
Rudimentary DBus toggle-to-talk support

kris7t:
#### Pre-flight Checklist

#### Description of Change
Adds a ToggleToTalk method to the DBus interface to unmute/mute the microphone in the active service if the recipe supports it.

We will need to add support for this feature in recipes.

#### Motivation and Context
Unmuting and muting the microphone by going into the application window can be tedious (e.g., while screen sharing during a meeting or while gaming), even if common courtesy dictates not running with an open mic in a noisy environment. A common solution for this is applications with global shortcuts that intercept a given keypress either as a push-to-talk (kept held while talking) or toggle-to-talk button.

Unfortunately, the security model of Wayland compositors on Linux disallows applications from intercepting global key presses. A common solution is to configure the key binding right in the compositor (which has the appropriate privileges to intercept it), and talk to the application over IPC. Since we already use DBus as IPC on Linux, this patch adds a new method to the DBus interface to support toggle-to-talk.

For example, this shell script can send a toggle-to-talk method call to Ferdium:

```bash
#!/usr/bin/env bash
exec dbus-send --session --dest=org.ferdium.Ferdium --type=method_call /org/ferdium org.ferdium.Ferdium.ToggleToTalk
```

I have bound it in the [sway](https://github.com/swaywm/sway) compositor like this:

```
bindsym $mod+z exec ~/.local/bin/toggle-to-talk.sh
```

Of course, this feature requires recipe support to appropriately handle the toggle-to-talk event. I deliberately went with a very basic solution to avoid managing mic state: on a toggle-to-talk method call, the active service is asked to toggle its mic. This means that if you switch away from a service during a call, it will no longer receive the toggle-to-talk events. However, the contrary (enable toggle-to-talk even for services in the background) would require us to track which call(s) are currently active, which would be a much more complicated and error-prone solution.

Currently, I haven't patched any recipes to support toggle-to-talk, but this doesn't look like a hard job. For example, Discord seems to work OK with

```javascript
Ferdium.toggleToTalk(() => document.querySelector('button[aria-label="Mute"]')?.click());
```

while Skype requires a slightly more complicated

```javascript
Ferdium.toggleToTalk(() => (document.querySelector('button[aria-label="Unmute"]') ?? document.querySelector('button[aria-label="Mute"]'))?.click());
```

These are all best effort: it might be the case that the mic currently can't be toggled at all. In that case, the toggle-to-talk event is just ignored.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

Added toggle-to-talk support over DBus

### DIFF
--- a/docs/dbus/org.ferdium.Ferdium.xml
+++ b/docs/dbus/org.ferdium.Ferdium.xml
@@ -38,6 +38,13 @@
     -->
     <method name="ToggleWindow" />
     <!--
+      ToggleToTalk:
+
+      Toggles the microphone in the active service if supported and currently
+      applicable.
+    -->
+    <method name="ToggleToTalk" />
+    <!--
       Muted:
 
       Gets or sets the current muted state of notifications.
--- a/src/lib/dbus/Ferdium.ts
+++ b/src/lib/dbus/Ferdium.ts
@@ -31,6 +31,10 @@ export default class Ferdium extends dbus.interface.Interface {
     this.dbus.trayIcon._toggleWindow();
   }
 
+  ToggleToTalk(): void {
+    this.dbus.trayIcon.mainWindow?.webContents.send('toggle-to-talk');
+  }
+
   emitUnreadChanged(): void {
     Ferdium.emitPropertiesChanged(
       this,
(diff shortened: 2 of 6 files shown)


mcmxcdev (approved):
Thanks @kris7t, looks really clean and simple!  :rocket:

PR #1555, opened 2024-02-02
Two Factor SMS token parser and copy to clipboard via notification.ts

dutscher:
#### Pre-flight Checklist

#### Description of Change

I have the daily trouble with two factor tokens they came via sms on my mobile phone. My ferdium is connected via android messages and the app shows every sms via ferdium notifications on my desktop. I wrote a little helper which parse the token out of the notification body and copy the found token into clipboard. Now i can press CTRL + V instead of look onto the notification and tip every digit into my input.

#### Motivation and Context

See description :D My motivation is my daily pain with tipping every digit or copy the tokens with accidentally copied whitespaces.

#### Checklist

- [X] `pnpm test` passes

#### Release Notes

Two Factor SMS Token is now copied to clipboard if you saw the + in your notification.

Thanks in advance

### DIFF
--- a/src/containers/settings/EditSettingsScreen.tsx
+++ b/src/containers/settings/EditSettingsScreen.tsx
@@ -99,6 +99,15 @@ const messages = defineMessages({
     id: 'settings.app.form.notifyTaskBarOnMessage',
     defaultMessage: 'Notify TaskBar/Dock on new message',
   },
+  twoFactorAutoCatcher: {
+    id: 'settings.app.form.twoFactorAutoCatcher',
+    defaultMessage: 'Auto catch two-factor codes from notifications',
+  },
+  twoFactorAutoCatcherArray: {
+    id: 'settings.app.form.twoFactorAutoCatcherArray',
+    defaultMessage:
+      'Comma seperated and case insensitive words/expressions to catch catch two-factor codes from. Ex.: token, code, sms, verify',
+  },
   navigationBarBehaviour: {
     id: 'settings.app.form.navigationBarBehaviour',
     defaultMessage: 'Navigation bar behaviour',
@@ -383,6 +392,8 @@ class EditSettingsScreen extends Component<
         privateNotifications: Boolean(settingsData.privateNotifications),
         clipboardNotifications: Boolean(settingsData.clipboardNotifications),
         notifyTaskBarOnMessage: Boolean(settingsData.notifyTaskBarOnMessage),
+        twoFactorAutoCatcher: Boolean(settingsData.twoFactorAutoCatcher),
+        twoFactorAutoCatcherArray: settingsData.twoFactorAutoCatcherArray,
         navigationBarBehaviour: settingsData.navigationBarBehaviour,
         webRTCIPHandlingPolicy: settingsData.webRTCIPHandlingPolicy,
         searchEngine: settingsData.searchEngine,
@@ -679,6 +690,23 @@ class EditSettingsScreen extends Component<
           ),
           default: DEFAULT_APP_SETTINGS.notifyTaskBarOnMessage,
           type: 'checkbox',
+        },
+        twoFactorAutoCatcher: {
+          label: intl.formatMessage(messages.twoFactorAutoCatcher),
+          value: ifUndefined<boolean>(
+            settings.all.app.twoFactorAutoCatcher,
+            DEFAULT_APP_SETTINGS.twoFactorAutoCatcher,
+          ),
+          default: DEFAULT_APP_SETTINGS.twoFactorAutoCatcher,
+          type: 'checkbox',
+        },
+        twoFactorAutoCatcherArray: {
+          label: intl.formatMessage(messages.twoFactorAutoCatcherArray),
+          value: ifUndefined<string>(
+            settings.all.app.twoFactorAutoCatcherArray,
+            DEFAULT_APP_SETTINGS.twoFactorAutoCatcherArray,
+          ),
+          default: DEFAULT_APP_SETTINGS.twoFactorAutoCatcherArray,
         },
         navigationBarBehaviour: {
           label: intl.formatMessage(messages.navigationBarBehaviour),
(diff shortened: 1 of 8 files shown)


SpecialAro:
Hi @dutscher ! Thank you for this nice addition 😄 

The code looks good and clean, thank you for that. I was just thinking that maybe we should add a toggle in the "Settings" under the "Privacy" or "Security" tab to Enable/Disable this feature given that some users might want to disable it.

Idk if you agree with this or not, let me know :D

SpecialAro (commented):
Can you also please fix the lint issues that are causing the CI builds to fail?

`pnpm prepare-code` should be enough to automatically fix all lint issues.

src/webview/notifications.ts
```diff
@@ -14,6 +14,27 @@ export class NotificationsHandler {
...
+      ---
+      483133 is your GitHub authentication code. @github.com #483133
+      ---
+      eBay: Ihr Sicherheitscode lautet 080090. \nEr läuft in 15 Minuten ab. Geben Sie den Code nicht an andere weiter.
+      ---
+      PayPal: Ihr Sicherheitscode lautet: 989605. Geben Sie diesen Code nicht weiter.
+      */
+      const rawBody = options.body;
+      const { 0: token } = /\d{5,6}/.exec(options.body) || [];
+      if (token && (['token', 'code'].find(a => options.body.toLowerCase().includes(a)))) {
```
vraravam: For some senders in India, they use the word `SMS` in the body

vraravam (changes requested):
Also - without the privacy settings as @SpecialAro mentioned, I would not want to merge this PR. As a security-conscious user, I do NOT want Ferdium to auto-parse SMSes and put that into the clipboard.

src/webview/notifications.ts
dutscher: i added "sms" to the array but maybe we make this also editable via settings?

dutscher:
thanks @SpecialAro and @vraravam 
eslint is fixxed. 

yes i thought about the same that there is a privacy setting. 
i tried to study your settings environment and found "settings.app.form.clipboardNotifications" which is similar the same to mine.
but i'm not able to implement it in this way as you did and i need your help to get this done.

what we need is:

In the "Privacy" tab a new toggle for: "Parse the notification content for a possible 2 factor token and copy to clipboard".
And maybe a input for the `['token', 'code', 'sms']` array, than that is also editable.

lets do that? i could help in the translation area for german!

thanks in advance :)

SpecialAro:
Hi @dutscher

I can fairly easily add a toggle to settings to enable/disable the feature when I get the time (maybe tomorrow). 

As per the array, what is the idea behind it? Leaving it as a simple inputbox where the user writes each word (coma separated) they want in order to be catched? Would that work?

For the German translation I wouldn't bother for now (only after we get this feature merged) because translations are handled by Crowdin, so we can change it there after merge to any language 😁 maybe you can help out there as well.

I'll get back to you on the toggle thing 😁

SpecialAro:
Hi @dutscher,

I just added the setting on the Settings > Privacy tab. Could you please debug and test if this works (I wasn't able to test it, only to code it). It should be fine but let me know if it works as intended.

The input box is a comma separated array of the keywords to catch the token. Also, if the user disables (it is disabled by default - is this how it should be @vraravam ?) then your code doesn't run - which is intended security wise.

vraravam:
yes, this should be an opt-in preference. So, off by default.

dutscher:
> Hi @dutscher,
> 
> I just added the setting on the Settings > Privacy tab. Could you please debug and test if this works (I wasn't able to test it, only to code it). It should be fine but let me know if it works as intended.
> 
> The input box is a comma separated array of the keywords to catch the token. Also, if the user disables (it is disabled by default - is this how it should be @vraravam ?) then your code doesn't run - which is intended security wise.

nice job @SpecialAro i try to test this on windows.

src/i18n/locales/en-US.json
```diff
@@ -260,6 +260,8 @@
   "settings.app.form.startMinimized": "Start minimized",
   "settings.app.form.translatorEngine": "Translator Engine",
   "settings.app.form.translatorLanguage": "Default Translator language",
+  "settings.app.form.twoFactorAutoCatcher": "Auto catch two-factor codes from notifications",
```
dutscher, line 263:
```suggestion
  "settings.app.form.twoFactorAutoCatcher": "Auto catch two-factor codes from notifications and copy to clipboard",
```

src/containers/settings/EditSettingsScreen.tsx
```diff
@@ -99,6 +99,15 @@ const messages = defineMessages({
     id: 'settings.app.form.notifyTaskBarOnMessage',
     defaultMessage: 'Notify TaskBar/Dock on new message',
   },
+  twoFactorAutoCatcher: {
+    id: 'settings.app.form.twoFactorAutoCatcher',
+    defaultMessage: 'Auto catch two-factor codes from notifications',
```
dutscher, line 104:
```suggestion
    defaultMessage: 'Auto catch two-factor codes from notifications and copy to clipboard',
```

dutscher:
# Default settings:
image

## Console output:
image

--

# Activated:
image
## Console output:
image

But i'm not able to test the notification. i will try later again

dutscher:
i test build and debug. but i wasn't able to test the window notification. therefore i create a codepen with the exact same code and some mocks on top: https://codepen.io/dutscher/pen/mdoGpvB?editors=0010
the settings works perfectly!
i change a little bit the settings description.

and i got some weird receipes changes into my pr: 
image

for me its done! 
do you have any open points?

cheers and thanks in advance

recipes
```diff
@@ -1 +1 @@
-Subproject commit 333c473afbe0504c26bc5fe89b607e98f8966be8
+Subproject commit f5b618f47b29597eb05e4c09ea664d827dcbda2c
```
SpecialAro, line 1: This should be left out

src/webview/notifications.ts
```diff
@@ -14,6 +14,52 @@ export class NotificationsHandler {
...
+
+        debug('wordsToCatch', wordsToCatch);
+
+        if (
+          token &&
+          wordsToCatch.some(a =>
+            options.body.toLowerCase().includes(a.toLowerCase()),
+          )
+        ) {
+          debug('Token parsed and copy to clipboard', token);
```
SpecialAro: Let's maybe not print this to the console? I'm thinking about security wise...

SpecialAro (changes requested):
LGTM but still haven't tested yet. Could you please fix my comments?

src/webview/notifications.ts
dutscher: done

PR #1580, opened 2024-02-17
chore: fix notification "option" clone

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Fix notifications object clone (options)

#### Motivation and Context
Several users are reporting that their services are not showing tray notifications (even though they ear the sound from the service receiving the notification).
After investigating using the Whatsapp service, I was able to see several errors in the console whenever a new notification came in:
image

This was a hint to investigate what was happening and I came into the conclusion that the `options` object was not being cloned properly (probable because services are injecting some function or keys that are not properly cloned). After doing a "hack" with `JSON.parse` and `JSON.stringify` (we do have this hack in several logic around the app), the `options` object is now properly cloned and doesn't throw any errors - and, therefore, enables users to get the message in the tray at the end.

I tested this with Whatsapp and it worked properly (fixed it actually), but several other services might benefit from this fix.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes
Fixed tray notifications for Whatsapp (and probably other services as well)

### DIFF
--- a/src/webview/recipe.ts
+++ b/src/webview/recipe.ts
@@ -124,8 +124,14 @@ contextBridge.exposeInMainWorld('ferdium', {
     safeParseInt(text),
   setDialogTitle: (title: string | null | undefined) =>
     dialogTitleHandler.setDialogTitle(title),
-  displayNotification: (title: string, options: any) =>
-    notificationsHandler.displayNotification(title, options),
+  displayNotification: (title: string, options: any) => {
+    notificationsHandler.displayNotification(
+      title,
+      // The following line is needed so that a proper clone of the "options" object is made.
+      // This line was causing issues with some services.
+      JSON.parse(JSON.stringify(options)),
+    );
+  },
   getDisplayMediaSelector,
 });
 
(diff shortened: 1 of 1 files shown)


vraravam (approved):

src/webview/recipe.ts
```diff
@@ -124,8 +124,14 @@ contextBridge.exposeInMainWorld('ferdium', {
...
   setDialogTitle: (title: string | null | undefined) =>
     dialogTitleHandler.setDialogTitle(title),
-  displayNotification: (title: string, options: any) =>
-    notificationsHandler.displayNotification(title, options),
+  displayNotification: (title: string, options: any) => {
+    notificationsHandler.displayNotification(
+      title,
+      // The following line is needed so that a proper clone of the "options" object is made.
+      // This line was causing issues with some services.
+      JSON.parse(JSON.stringify(options)),
```
vraravam: imo, it might be a good idea to move this fix into the jsUtils file and add a test for the same. wdyt? (especially if its repeated in multiple places in the codebase)

SpecialAro: Totally agree!

Just noticed we already have a `cleanseJSObject` util for that. Just changed the code.

Thank you @vraravam !

vraravam (approved):

PR #1740, opened 2024-05-04
chore: change icons to reduce bundle

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
Reduce user bundle on recipes by removing SVG local reference

#### Motivation and Context
After some discussion with @vraravam we came to the conclusion that SVG increase local bundle size for users, which is unnecessary given users can fetch the static assets generated by `jsdelivr`.

My idea is that we can still have some control over the icons by making use `jsdelivr` cdn to point to GH assets for the `icon.svg` of each recipe, whilst giving the possibility for users to create (or modify existing recipes) by making use of `defaultIcon` key (that should contain an URL or be undefined - more in the twin PR here: )

To test this PR do the following:
- Checkout this PR.
- Go to the `recipes` submodule and add my `ferdium-recipes` fork as a remote
- Checkout the branch PR here: https://github.com/ferdium/ferdium-recipes/pull/537
- Run `pnpm i` and `pnpm package`
- Go back to the main repo folder
- Compile the app and run it for the first time (add 2 or 3 recipes)
- Go to your system folder (in case of Windows `%AppData%/Ferdium/recipes`) and manually bring the version down of each recipe `package.json`
- Reload Ferdium and wait for the prompt to reload services.

What to expect:
- Icons won't change (given we are using the same icons)
- The SVG file in the `%AppData%/Ferdium/recipes/{recipeId}` folder should be automatically deleted.

This PR needs to be merged AFTER the one on the recipes repo (https://github.com/ferdium/ferdium-recipes/pull/537).

This PR was tested in:
- [x] Windows 11
- [x] MacOS
- [ ] Linux

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/src/api/server/ServerApi.ts
+++ b/src/api/server/ServerApi.ts
@@ -12,6 +12,7 @@ import {
   statSync,
   writeFileSync,
 } from 'fs-extra';
+import ms from 'ms';
 import tar from 'tar';
 
 import RecipeModel, { type IRecipe } from '../../models/Recipe';
@@ -460,7 +461,7 @@ export default class ServerApi {
     }
     debug(archivePath);
 
-    await sleep(10);
+    await sleep(ms('10ms'));
 
     await tar.x({
       file: archivePath,
@@ -471,7 +472,7 @@ export default class ServerApi {
       onwarn: x => debug('warn', recipeId, x),
     });
 
-    await sleep(10);
+    await sleep(ms('10ms'));
 
     const { id, defaultIcon } = readJsonSync(
       join(recipeTempDirectory, 'package.json'),
--- a/src/api/server/ServerApi.ts
+++ b/src/api/server/ServerApi.ts
@@ -473,11 +473,19 @@ export default class ServerApi {
 
     await sleep(10);
 
-    const { id } = readJsonSync(join(recipeTempDirectory, 'package.json'));
+    const { id, defaultIcon } = readJsonSync(
+      join(recipeTempDirectory, 'package.json'),
+    );
     const recipeDirectory = join(recipesDirectory, id);
     copySync(recipeTempDirectory, recipeDirectory);
     removeSync(recipeTempDirectory);
     removeSync(join(recipesDirectory, recipeId, 'recipe.tar.gz'));
+
+    // TODO: This is a temporary fix to remove svg icons from the user AppData. This should be removed after some versions.
+    debug('Removing default icon', defaultIcon);
+    if (defaultIcon) {
+      removeSync(join(recipeDirectory, 'icon.svg'));
+    }
 
     return id;
   }
--- a/src/api/server/ServerApi.ts
+++ b/src/api/server/ServerApi.ts
@@ -481,8 +481,7 @@ export default class ServerApi {
     removeSync(recipeTempDirectory);
     removeSync(join(recipesDirectory, recipeId, 'recipe.tar.gz'));
 
-    // TODO: This is a temporary fix to remove svg icons from the user AppData. This should be removed after some versions.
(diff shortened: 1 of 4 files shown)


vraravam:
@SpecialAro  - after building using both the main app and the recipes (your branches), I still don't see the `icon.svg` to be deleted in the installation profile folder.

SpecialAro:
> @SpecialAro  - after building using both the main app and the recipes (your branches), I still don't see the `icon.svg` to be deleted in the installation profile folder.

Huum... That is weird. Can you check what you get in this debug?

```
    debug('Removing default icon', defaultIcon); 
     if (defaultIcon) { 
       removeSync(join(recipeDirectory, 'icon.svg')); 
     } 
 ```

**Comment by vraravam** [comment]
for pre-existing recipes in a pre-existing profile, the `icon.svg` does not get auto-deleted UNLESS there's a version bump in the recipe. For eg, I tried for the gmail service (installed recipe), and the `icon.svg` remained until I bumped up the version of gmail recipe and repackaged. Then, Ferdium replaces the whole folder and prompts to "Reload services"

Update: This is not such a big deal. In fact, if I forcibly remove the `icon.svg` file without a version bump, then some other  (db?) reference is still present to the non-existent file on the local file-system, which then makes the sidebar show up **without** any icons. So, probably this auto-delete not working is a boon in disguise!

**Comment by SpecialAro** [comment]
Oh! Maybe I wasn't clear enough, sorry. I meant to say that SVGs are only updated **after** a version bump (that is why I said to locally bump down the version on your profile - to simulate an update).

I purposely added code (and didn't remove it) to we ensure that all users still get a functional Ferdium with icons, without a breaking change. This way, after merging, we can trigger a version bump on all recipes (we can make a PR just for bumping all recipes) and, consequently, users will see their SVGs gone, but will have the new versions with `defaultIcons` defined.

Does that make sense? Do you agree with this?

Note: one thing that I didn't test/accounted for is the custom recipes. But I think that, for those, we can keep the possibility of having a local SVG (so we would remove the TODO comment and leave the code as permanent).

**Comment by vraravam** [comment]
yes, i agree. if needed, we can do the version-bump for all recipes.

**Review by vraravam — APPROVED** [review]
LGTM! 
Nice work @SpecialAro

**Comment by SpecialAro** [comment]
Thank you very much @vraravam!

Should we also test on Linux before merging (and also remove the debug statement and the TODO comment?)

(I'm heading off, but I can do that tomorrow) 😁

**Comment by vraravam** [comment]
> Should we also test on Linux before merging (and also remove the debug statement and the TODO comment?)

i dont think there's any OS-specific code in there. So, it should work in linux too.

I can remove the debug statements and the TODO comments now

File: src/api/server/ServerApi.ts [hunk 1]
```diff
@@ -481,8 +481,7 @@ export default class ServerApi {
     removeSync(recipeTempDirectory);
     removeSync(join(recipesDirectory, recipeId, 'recipe.tar.gz'));
 
-    // TODO: This is a temporary fix to remove svg icons from the user AppData. This should be removed after some versions.
-    debug('Removing default icon', defaultIcon);
+    // TODO: This is a temporary fix to remove svg icons from the user AppData. This should be removed after versions of all recipes have been bumped up
```
SpecialAro, line 485: @vraravam in reality this code should **not** be removed to ensure that user local recipes (custom recipes made by users) don't lose functionality. If we keep the next line we ensure that everything works properly 😁

I propose we delete this TODO comment as we have nothing to do 🙏

vraravam: yes, i didn't think of custom recipes. We can remove it in the next PR whenever that happens

PR #1924, opened 2024-10-12
Upgrade electron-builder, electron-updater and electron/notarize

Alphrag:
#### Pre-flight Checklist

#### Description of Change

Upgrade the `electron-builder` and `electron-updater` to the latest version, as well as `@electron/notarize`.

#### Motivation and Context

The version of `electron-builder` and `electron-updater` have not been updated in a while since it was breaking the builds when rebuilding the `sqlite` and/or `macos-notifcation-state` dependencies. The `alpha` version finally bumps the `@electron/rebuild` which contains a newer version of `gyp` fixing the problem encountered. This should also fix the correct upload to the snap store, and improve the auto-update sequences on Windows (by preventing concurrent downloads), as well as improving it for Linux targets.

Additionally, In line with the current dependency of `electron-builder`, this upgrade `notarize` and make some changes to push all the required values for the notarization process inside environment variables, to mimic its usage in electron-builder.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

Upgrade `electron-builder`,`electron-updater` and `@electron/notarize` to the latest version.


package.json
```diff
@@ -81,7 +81,7 @@
     "electron-dl": "3.5.2",
     "electron-find": "1.0.7",
     "electron-react-titlebar": "1.2.1",
-    "electron-updater": "6.2.1",
+    "electron-updater": "6.4.0-alpha.0",
```
vraravam, line 84: @Alphrag  - should we upgrade to an alpha version? Would that be stable for our needs?

package.json
```diff
@@ -164,7 +164,7 @@
     "concurrently": "8.2.2",
     "cross-env": "7.0.3",
     "electron": "32.2.0",
-    "electron-builder": "24.13.3",
+    "electron-builder": "26.0.0-alpha.1",
```
vraravam, line 167: same concern about alpha version

package.json
Alphrag: For `electron-updater`, the only commit on the alpha version is to do with the auto-update of pacman, so we can go to the stable version if you prefer, but that should not bother us. The only reason I decided to put this version, is to match with the electron-builder introducing the pacman possible update, so should we want to add it to our possible targets, they would work straight away together without having to figure out whether one of them needs updating first.

package.json
Alphrag: For `electron-builder`, I would be really enclined to use that version even if it is an alpha version for a few different reasons that go together:
- this version updates the `electron/rebuild` dependency to a version that uses `gyp>=10`, whereas the latest stable release of electron-builder does not have it, and this causes the build to fail at the step when rebuilding `sqlite` and `macos-notification-state` on arch different from the current arch used;
- if we were to use the stable version, we would need to make a special step to install the python packages like `distutils` which are no longer included by default since python3.12 but are necessary by the rebuild process with the version of `electron/rebuild` included in `electron-builder v5.1.8`;
- the last time `electron-builder` went into alpha release mode, it took 4 months to get out of it, so it would mean waiting for a while to get this very useful update;
- we are only at the beginning of the alpha release mode, which is planned to bring big changes but so far, (in `26.0.0-alpha.0` qnd `26.0.0-alpha.1`) only relatively small updates have been pushed, with my only concern regarding the change in https://github.com/electron-userland/electron-builder/pull/8576 which modifies the handling of including node_modules in the app, so I think this does not affects us.

Alphrag: So if you think it is better to stick with stable release, I can try to modify the build script in order to make sure that we add the python packages, even though I am not a big fan of this. Mostly because it would require the developers to know about it and to have an environment with the actual python containing the required packages when building with local build scripts if they do some cross-arch compiling.

vraravam: if the alpha is only for the introduction of pacman, then i think we should be ok.

Since this PR also upgrades the notarization aspect (which is only used for builds from GHA), could you please ensure to monitor the upcoming nightly builds a couple of times?

vraravam (approved):

PR #1927, opened 2024-10-14
Revert "Upgrade electron-builder, electron-updater"

Alphrag:
Partially reverts ferdium/ferdium-app#1924

## Reason for the revert
Despite the app building and launching from a clean state, it appears that there is then a problem when the internal server starts, with some dependencies missing. In particular, the module `color-string` needed by `adonisjs` is not included, as can be seen from the inspection of the `.asar` file: (where the top tree correspond to the asar when using `electron-builder` version `26.0.0-alpha`, and the bottom one to what we have in the working version built with version `24.13.3`)
Capture d’écran 2024-10-14 à 18 32 28

Despite multiple attempts to fix this, the problem is deeper than expected. We are reverting the upgrade of `electron-builder` and `electron-updater` in order to avoid breaking the app on the next triggered nightly.

## Changes kept
The upgrade of `@electron/notarize` together with the passing of the `teamId` value necessary during notarization into an environment variable.


Alphrag:
Superseded by #1928

PR #1934, opened 2024-10-20
Upgrade esbuild to 0.24.0

vraravam:
#### Pre-flight Checklist

#### Description of Change

Upgrade esbuild to the latest version with accompanying changes in the APIs

#### Motivation and Context

Upgrade outdated packages.

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/esbuild.mjs
+++ b/esbuild.mjs
@@ -79,7 +79,6 @@ const runEsbuild = async () => {
   const startTime = performance.now();
 
   const myArgs = process.argv.slice(2);
-  const isDev = myArgs.includes('--watch');
   log(chalk.blue('Starting with args'), myArgs);
 
   if (fs.existsSync(outDir)) {
@@ -99,7 +98,7 @@ const runEsbuild = async () => {
   );
 
   // Run build
-  await esbuild.build({
+  const context = await esbuild.context({
     entryPoints,
     format: 'cjs',
     minify: false,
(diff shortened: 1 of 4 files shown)


vraravam:
@vraravam has tested on mac on m1 (build the app and run it as a standalone app (not as a debug/dev mode with hot-reload - which stopped working for me approx 1 year ago)

Alphrag:
The build script doesn't work because of the `tsc` call in the `pnpm prepare-code` (this made me realise that this is not done in the gh action build), but the rest goes through fine.
Capture d’écran 2024-10-20 à 14 30 34

vraravam:
> The build script doesn't work because of the `tsc` call in the `pnpm prepare-code` (this made me realise that this is not done in the gh action build), but the rest goes through fine.

I don't think this is the case, since `prepare-code` is being called [here](https://github.com/ferdium/ferdium-app/blob/develop/.github/workflows/builds.yml#L95)

Also, on my local, I'm able to get a fully green build ie which results in a zero exit code when invoking the `scripts/build-unix.sh` script. If you are testing on windows, is that script out of date?

Alphrag:
Oh you're right, I was looking for it next to the `lint` and `test`, and didn't think of looking there. 
No, I am testing it on mac, with a `CLEAN=true ./scripts/build-unix.sh` and got the same error on both x64 and M3.

PR #1944, opened 2024-10-24
style(tabs): improve visual consistency and label handling for horizontal style

DenysMb:
#### Pre-flight Checklist

#### Description of Change
When we move the sidebar to the top and show the services label, we get a very inconsistent visual.
The service tab width is different according to the size of the service label and the active indicator pushes everything down, so the active service tab keeps unaligned with all the others.

In this PR I fixed both issues by using a fixed width for all the tabs when labels are displayed and using box-shadow instead of a border for the active tab indicator so things are not pushed down, This way we have a more consistent look for all the tabs and the whole application in general.

Another problem is that the long labels were not "ellipsed, I fixed this too.

#### Motivation and Context
I prefer to use the tabs at the top of the application, but all these little issues prevented me from doing so because they bothered me a lot.

#### Screenshots
Before:
Screenshot_20241024_181046
Screenshot_20241024_181052

After:
Screenshot_20241024_181224

#### Checklist

- [X] `pnpm test` passes

#### Release Notes
- All tabs, when the service label is enabled and the sidebar is on the top, now have the same width, labels get ellipsed and the activated tab is no longer un-aligned with others.

### DIFF
--- a/src/features/appearance/index.ts
+++ b/src/features/appearance/index.ts
@@ -248,7 +248,6 @@ const generateServiceRibbonWidthStyle = (
       ${useGrayscaleServices ? graysacleServices : null},
     }
     .tab-item .tab-item__label {
-      margin-left: 4px !important;
       font-size: ${fontSize}px !important;
     }
     .tab-item.is-label-enabled {
--- a/src/features/appearance/index.ts
+++ b/src/features/appearance/index.ts
@@ -147,7 +147,7 @@ const generateServiceRibbonWidthStyle = (
 ) => {
   const width = Number(widthStr);
   const iconSize = Number(iconSizeStr) - iconSizeBias;
-  const tabItemWidthBias = 3;
+  const tabItemWidthBias = 1;
   const verticalStyleOffset = 29;
 
   let fontSize: number;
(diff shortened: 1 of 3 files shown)


vraravam (approved):
LGTM

vraravam:
@all-contributors please add @DenysMb for code

wvds:
@DenysMb @vraravam I think this PR breaks the use of the accent color for the tabs.
Before the change, the accent color was used for the "active" state of the tab, but with this change it's always purple (the default accent color of Ferdium).

DenysMb:
> @DenysMb @vraravam I think this PR breaks the use of the accent color for the tabs. Before the change, the accent color was used for the "active" state of the tab, but with this change it's always purple (the default accent color of Ferdium).

Very strange, since I am using the SCSS $theme-brand-primary variable that should handle this.
I'll take a look now to see what broke this.

**Edit**: I was wrong. The accent color is controlled in the `appearance/index.ts`. I changed from border to box-shadow in SCSS but there still changes the border-color. So, I just need to change there from border-color to box-shadow too and it will work. I'll open a PR for this.

DenysMb:
@wvds I just opened a PR to fix this: https://github.com/ferdium/ferdium-app/pull/1967
Thanks for letting me know. Nobody wants their contribution to cause a bug.

PR #1545, opened 2024-01-24
feat: self signed certificates bypass

SpecialAro:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change
Feature: add possibility to add and enable self-signed certificates

#### Motivation and Context
This fixes https://github.com/ferdium/ferdium-app/issues/1227 by adding a global possibility of enabling Self-Signed Certificates (on Settings > Advanced). Only certificates that match the certificates in the specified folder enable the user to bypass the `certificate-error` event.

#### Screenshots
image
image

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (`pnpm prepare-code`)
- [x] `pnpm test` passes
- [x] I tested/previewed my changes locally

#### Release Notes
feat: self signed certificates bypass

PR #2054, opened 2025-02-27
support for Custom Root CA

tomaszduda23:
#### Pre-flight Checklist

#### Description of Change

Follow up on https://github.com/ferdium/ferdium-app/pull/1545. Using per host certificates is a little bit inconvenienced. This PR add support for Custom Root CA.

#### Motivation and Context

Validation time of host certificates can be very short (a few months). It is easier to use Custom Root CA which is usually valid for a few years.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

support for Custom Root CA

### DIFF
--- a/src/helpers/certs-helpers.ts
+++ b/src/helpers/certs-helpers.ts
@@ -1,33 +1,33 @@
 import { X509Certificate } from 'node:crypto';
 import { join } from 'node:path';
+import type { Certificate } from 'electron';
 import { ensureDirSync, readFileSync, readdirSync } from 'fs-extra';
 import { userDataCertsPath } from '../environment-remote';
 import { removeNewLines } from '../jsUtils';
 
-export const checkIfCertIsPresent = (clientCert): boolean => {
+export const checkIfCertIsPresent = (clientCert: Certificate): boolean => {
   const certsFolder = userDataCertsPath();
 
   ensureDirSync(certsFolder);
 
   const certs: string[] = [];
 
-  for (const file of readdirSync(certsFolder)) {
-    const cert = readFileSync(join(certsFolder, file), {
-      encoding: 'utf8',
-      flag: 'r',
-    });
-    try {
+  try {
+    const certToVerify = new X509Certificate(clientCert.issuerCert.data);
+
+    for (const file of readdirSync(certsFolder)) {
+      const cert = readFileSync(join(certsFolder, file), {
+        encoding: 'utf8',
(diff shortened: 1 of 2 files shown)


Alphrag:
Hi @tomaszduda23, thanks for this PR.
Even though this seems all good, I am however a bit reluctant to merge this as it is, mostly because that would mean adding a new dependency which has not been updated in 3 years and has even older dependencies itself. We are trying to slowly get rid of packages that are no longer updating, so introducing a new one does not seem the path forward. 
Is there no other module that would allow us to verify certificates, like using the [`pkijs` module](https://www.npmjs.com/package/pkijs), or directly the [Crypto api of node](https://nodejs.org/docs/latest-v22.x/api/crypto.html)?

tomaszduda23:
@Alphrag good point. I updated to use crypto api of node.

src/helpers/certs-helpers.ts
```diff
@@ -1,9 +1,10 @@
+import { X509Certificate } from 'node:crypto';
 import { join } from 'node:path';
 import { ensureDirSync, readFileSync, readdirSync } from 'fs-extra';
 import { userDataCertsPath } from '../environment-remote';
 import { removeNewLines } from '../jsUtils';
 
-export const checkIfCertIsPresent = (certData: string): boolean => {
+export const checkIfCertIsPresent = (clientCert): boolean => {
```
Alphrag: Can you please add the type of the parameter? It should be of type `Certificate` from `electron`.

src/helpers/certs-helpers.ts
```diff
@@ -15,8 +16,19 @@ export const checkIfCertIsPresent = (certData: string): boolean => {
       encoding: 'utf8',
       flag: 'r',
     });
+    try {
+      const caCert = new X509Certificate(cert);
+      if (caCert.ca) {
+        const certToVerify = new X509Certificate(clientCert.issuerCert.data);
```
Alphrag: Can you also put this variable out of the `for` loop since we can define it once and for all?

Alphrag (commented):
Thank you very much for changing your implementation with using node api. It is quite nice that way! 
Can you please make the two changes I suggest, and then it will be good for merging.

tomaszduda23:
Updated according to comments.

Alphrag:
Thanks @tomaszduda23 for this PR!
@all-contributors please add @tomaszduda23 for code

PR #1920, opened 2024-10-10
Allow customising shortcuts for activating next/previous service, see #53

insidewhy:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change
Allow customising certain shortcuts, for now just "activate next service" and "activate previous service".

Also changed default service switch shortcuts from `Cmd+Tab`/`Cmd+Shift+Tab` to `Ctrl+Tab`/`Ctrl+Shift+Tab` as the former key bindings are already bound by the OS and don't work, see https://github.com/ferdium/ferdium-app/pull/1920#issuecomment-2407203236

#### Motivation and Context

Ctrl+tab and ctrl+shift+tab are not so ergonomic and this is an action that I use very frequently. This has been requested. Many people complained about these specific keyboard shortcuts in https://github.com/ferdium/ferdium-app/issues/53 and personally I prefer to use ctrl+, and ctrl+. for this.

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (`pnpm prepare-code`)
- [x] `pnpm test` passes
- [x] I tested/previewed my changes locally

#### Release Notes

Allow customising the keyboard shortcuts used to activate the next and previous services.

PR #2072, opened 2025-03-11
fix: give default value to shortcuts accelerators in menu

Alphrag:
#### Pre-flight Checklist

#### Description of Change

Give a default value to the shortcuts accelerators in the menu so that it does not crash if the store has not yet updated with the new settings. This comes from the possibility to have custom shortcuts as implemented in #1920.

#### Motivation and Context

As reported by some user in the Discord, it seems that for some people on Windows 10, the update that allowed custom shortcuts cause a complete unusable app. This potentially relates to a startup race condition occurring when the store is not updated with the correct settings of the shortcuts when the application starts, thus causing an issue when trying to create the menu if it is undefined.

Note: I was not able to reproduce the initial error, so even though this seems to work fine on my computers, I cannot make sure that it fully fixes the problem.

#### Checklist

- [x] `pnpm test` passes

### DIFF
--- a/src/lib/Menu.ts
+++ b/src/lib/Menu.ts
@@ -18,7 +18,11 @@ import type { StoresProps } from '../@types/ferdium-components.types';
 import { importExportURL, serverBase, serverName } from '../api/apiBase';
 // @ts-expect-error Cannot find module '../buildInfo.json' or its corresponding type declarations.
 import { gitBranch, gitHashShort, timestamp } from '../buildInfo.json';
-import { CUSTOM_WEBSITE_RECIPE_ID, LIVE_API_FERDIUM_WEBSITE } from '../config';
+import {
+  CUSTOM_WEBSITE_RECIPE_ID,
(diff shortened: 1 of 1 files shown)


Alphrag:
@SpecialAro I think that this should fix the problem introduced by the initial PR as it seems to me to be the most likely place where something could go wrong, but feel free to change other things if you think that the culprit is more at initialisation time.

SpecialAro (approved):
This looks good to me. Have we tested with the user on Discord to see if it works on its setup (as we are not able to reproduce the issue ourselves)?

Alphrag:
@SpecialAro Since we have now seen that the problem encountered by some people was not from this, shall we still merge this change for good measure even though it doesn't bring much?

SpecialAro:
> @SpecialAro Since we have now seen that the problem encountered by some people was not from this, shall we still merge this change for good measure even though it doesn't bring much?

Yes I think we shall, because it improves code nevertheless

PR #2076, opened 2025-03-12
Fix issue with accent color

SpecialAro:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change
Fix issue with accent color

#### Motivation and Context
Fixes https://github.com/ferdium/ferdium-app/issues/2017
Fixes https://github.com/ferdium/ferdium-app/issues/2047
Users who got the settings.json corrupted from this need to manually change the `accentColor` and the `progressbarAccentColor` to a random color (default: `#7367F0`)

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (`pnpm prepare-code`)
- [x] `pnpm test` passes
- [x] I tested/previewed my changes locally

PR #2091, opened 2025-03-27
Improve verification of accent colours

Alphrag:
#### Pre-flight Checklist

#### Description of Change

This improves the verification of the input when a user changes the accent color settings, by using the `tinycolor` module and introducing some fail-safe checks.

#### Motivation and Context

In #2076 some minimal verification was introduced to make sure that the a user can change the `Accent Colors` of the app manually by inserting a value, but it has some limitations:
- if the first character is not `#`, then it breaks the renderer, making the app unusable, and the only 2 ways to fix that are to either use the developer tools with a carefully crafted command, or to manually update this value in the settings file before relaunching the app;
- if the input can be correctly parsed but does not correspond to a colour, then any other modification of the settings will not work, until an appropriate colour value is written;
- named colours are not available at all, while colours given like `rgb(24,87,255)` are valid but must be pasted directly (or the parenthesis must be written first).

This PR addresses these problems by doing the following:
- if the input of one of the colours is empty, then its value is set to the default Ferdium colour;
- use the module `tinycolor` to verify the colours and deal with conversions;
- when a colour is modified, verify if the input is a valid colour, and submit the form to update the settings when this is the case, otherwise preventing the form to be submitted;
- if a user has modified a colour field to a value that is not valid, and changes another setting, then then colour field is reset to the last compatible value recorded in the settings configuration file before updating the settings. This way, even if the keeps an invalid input for one of the colour, it won't impede his capacity to change other settings;
- allow the usage of named colours, by converting them into their hex value so that `react-color` accepts them.

Know problems remaining:
- if the user enters a wrong input, then uses the slider, then the two will become disconnected and the accent colour will be black. In order to get the ability to use the slider again, the user must either enter a correct value or clear the input.

#### Screen recordings

Current behaviour:

https://github.com/user-attachments/assets/86c45880-7395-4491-b6de-3f68b4aa90c4

New behaviour obtained by this PR:

https://github.com/user-attachments/assets/83a63dfe-c1c7-44e7-b306-b73dc3073152

#### Checklist

- [x] `pnpm test` passes


Alphrag:
@SpecialAro I would have liked to make the verifications directly into the `ColorPickerInput` component. However, I have trouble figuring out how to do so because of the peculiar behaviour of the `onChange` and `onColorChange` methods of that class. I'm open to suggestions if you have a better way to deal with that!

PR #2093, opened 2025-03-28
chore: fix husky shebang deprecation warning

nathanaelhoun:
#### Pre-flight Checklist

#### Description of Change

Remove the shebang to follow changelog notes: https://github.com/typicode/husky/releases/tag/v9.1.2

Use the presence of node_modules to check if the commands should be run or not

#### Motivation and Context

Remove the following warning when committing

```
husky - DEPRECATED

Please remove the following two lines from .husky/pre-commit:

#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

They WILL FAIL in v10.0.0
```

#### Screenshots

/

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

none


.husky/commit-msg
```diff
@@ -1,10 +1,8 @@
...
-if [ -f $FILE_NAME ]; then
-  . $FILE_NAME
-
-  npx --no-install commitlint --edit "$1"
+if ! [ -d node_modules ]; then
+  echo "Skipping $0 hook as node modules are not installed" >&2
+  exit 0
 fi
+
+echo "" # empty line to help reading
```
Alphrag: Since there are lots of line being written down by the test output, it might be worth printing something more explicit to show what the following output is about. Since each step happening during `pnpm prepare-code` &co is displayed by showing two empty lines followed by a line starting with `> `, I suggest we do the same here to announce this step and don't just print an empty line, giving something like 
```


> checking commit message with `commitlint`
```
or similar. What do you think?

Alphrag (commented):
Thanks to your PR, I have now realised that my commit hooks have never been running correctly because the software I use to manage the git structure of my repos was not properly set, so it was directly skipping them.

.husky/commit-msg
nathanaelhoun:
```suggestion

echo ""
echo "> checking commit message with \`commitlint\`"
```

It's okay for me. I only added one newline because the first one is already created (I assume by as a space after the last command).

Alphrag (approved):

PR #2261, opened 2025-10-03
Fix Google SSO/SAML authentication; fixes #1973

### LINKED ISSUE #1973: Google Mail not loading (Company login)
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain (older issues were there, but nothing current)

### Ferdium Version

7.0.0

### What Operating System are you using?

Windows

### Operating System Version

Edition	Windows 10 Pro Version	22H2 Installed on	‎06/‎08/‎2022 OS build	19045.5131 Experience	Windows Feature Experience Pack 1000.19060.1000.0

### What arch are you using?

x64

### Last Known Working Ferdium version

7.0.0 (it just stopped working suddenly, without updating to a newer Ferdium version, so thats why this is same as above)

### Expected Behavior

Service loads and i can read my emails

### Actual Behavior

After logging in to my account, there is a short loading animation (something rotating, and google being written). But afterwards, everything is grey. There are no errors in the dev console during this process.

before logging in (aka after doing ctrl R to refresh the service):

image

after logging in:

image

after finish rotating:

image

I have already tried setting user agent to `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36` but it doesn't work.

Gmail still works fine for my personal gmail account, but not for the one at work.

### Steps to reproduce

I don't know if this can be reproduced easily with just any login, as for myself it still works fine with my personal account.

But the steps are:

1. add Gmail service
2. login with company account
3. nothing is displayed after logging in

Additional note: after doing ctrl R to reload the service, i have to relogin.

### Debug link

_No response_

### Screenshots

_No response_

### Additional information

_No response_

mbentley:
#### Pre-flight Checklist

#### Description of Change
Skip the reload because it will cancel POST requests which breaks SSO/SAML auth

#### Motivation and Context
This fixes #1973 

#### Screenshots
n/a

#### Checklist

- [X] `pnpm test` passes

#### Release Notes

Fixes SSO/SAML authentication issues

### DIFF
--- a/src/models/UserAgent.ts
+++ b/src/models/UserAgent.ts
@@ -14,8 +14,6 @@ export default class UserAgent {
 
   @observable.ref webview: ElectronWebView = null;
 
-  @observable chromelessUserAgent: boolean = false;
-
   @observable userAgentPref: string | null = null;
 
   @observable overrideUserAgent = (): string => '';
@@ -70,39 +68,35 @@ export default class UserAgent {
   }
 
   @computed get userAgent(): string {
-    return (
-      this.serviceUserAgentPref ||
-      (this.chromelessUserAgent
-        ? this.userAgentWithoutChromeVersion
-        : this.defaultUserAgent)
-    );
+    return this.serviceUserAgentPref || this.defaultUserAgent;
   }
 
   @action setWebviewReference(webview: ElectronWebView): void {
     this.webview = webview;
   }
 
-  @action _handleNavigate(url: string, forwardingHack: boolean = false): void {
+  @action _handleNavigate(url: string): void {
     if (url.startsWith('https://accounts.google.com')) {
-      if (!this.chromelessUserAgent) {
-        debug('Setting user agent to chromeless for url', url);
-        this.chromelessUserAgent = true;
-        this.webview.userAgent = this.userAgent;
-        if (forwardingHack) {
-          this.webview.loadURL(url);
-        }
-      }
-    } else if (this.chromelessUserAgent) {
-      debug('Setting user agent to contain chrome for url', url);
-      this.chromelessUserAgent = false;
-      this.webview.userAgent = this.userAgent;
+      debug('Setting user agent to chromeless for url', url);
+      // Set chromeless user agent (without Chrome version) for Google accounts
+      this.webview.userAgent =
+        this.serviceUserAgentPref || this.userAgentWithoutChromeVersion;
+    } else {
+      debug('Setting user agent to default for url', url);
+      // Set default user agent for all other sites
+      this.webview.userAgent =
+        this.serviceUserAgentPref || this.defaultUserAgent;
     }
+    // Note: We don't reload the URL here (previously done with loadURL() on will-navigate)
+    // because it cancels POST requests, which breaks SSO/SAML authentication flows
+    // (e.g., ACS endpoint requests). The user agent change takes effect on the
+    // current navigation without needing a reload.
   }
 
   _addWebviewEvents(webview: ElectronWebView): void {
     debug('Adding event handlers');
 
-    this._willNavigateListener = event => this._handleNavigate(event.url, true);
+    this._willNavigateListener = event => this._handleNavigate(event.url);
     webview.addEventListener('will-navigate', this._willNavigateListener);
 
     this._didNavigateListener = event => this._handleNavigate(event.url);


src/models/UserAgent.ts
```diff
@@ -82,15 +82,16 @@ export default class UserAgent {
     this.webview = webview;
   }
 
-  @action _handleNavigate(url: string, forwardingHack: boolean = false): void {
+  @action _handleNavigate(url: string): void {
     if (url.startsWith('https://accounts.google.com')) {
       if (!this.chromelessUserAgent) {
```
featherbear: In the Ferdium codebase, `chromelessUserAgent` seems to exist only in this file.
`chromelessUserAgent` semantically means a Chrome user agent without the version number.

Wondering if we even need this variable or conditional logic - it cuts down on the number of reassignments made to `this.webview.userAgent`, but I think the effects to performance is negligible compared to the complexity it adds.

Otherwise lgtm, can validate the fixes work for me too

featherbear: * For `account.google.com`
  * `this.webview.userAgent` will always be `this.serviceUserAgentPref || this.userAgentWithoutChromeVersion`
* For every other site, `this.serviceUserAgentPref || this.defaultUserAgent`

mbentley:
I posted something similar in the issue and forgot to include it when I submitted the PR- I have no idea if this will break other stuff and in full transparency, I used Claude Code to figure out why this was happening.

I took a look at the code was originally added in https://github.com/ferdium/ferdium-app/commit/2ad39ffb1cb0d0e5f79d6948f798ca79ed73c76c just to try to understand why it was added in the first place. I did test to see if TickTick authentication still works and it does so unless it was more than the auth, it seems to be good.

I had Claude do some analysis of why this shouldn't break things (as if the excessive emoji use didn’t give it away).

What that commit did:
Added chromeless user agent for Google login (removes Chrome version)
Added the forwardingHack parameter with loadURL() to reload the page
This was needed because TickTick requires Chrome version, but Google blocks it
The problem with the original approach:
The loadURL() reload was a workaround to apply the user agent change
It worked for simple GET requests
But it breaks POST requests (like SAML SSO) by canceling and reloading them as GET
Why Our Fix is Safe
What changed in Electron since 2021: The commit is from June 2021. Modern Electron (which Ferdium now uses) applies user agent changes immediately to ongoing navigation without needing a reload. Our fix:
✅ Still sets this.webview.userAgent = this.userAgent (line 90)
✅ Chromeless user agent still applies to Google accounts
✅ TickTick still gets the chromeless user agent it needs
❌ No more loadURL() that cancels POST requests
✅ SSO/SAML now works (as you confirmed!)

If further changes need to be made, I can take a look but this honestly isn’t my area of expertise.

mbentley:
I took at shot at simplifying things.  I left it as it's own commit for now in case it needs to be backed out.

Alphrag:
@mbentley Thanks a lot for looking into this and raising this PR. At first glance it seems very legitimate, but I would need to have a closer look at it to check it it could have side effects. In the meantime, can you just make sure to run the build+lint locally and push the fixes so that the builds can work through it?

@kris7t If you have a bit of time, I'd be happy to have your input on why was this implemented in the first place (since it seems this was you who made this), and if you think such a logic is still necessary? Thanks

mbentley:
Ah sorry about that, I was working from my laptop without all of the required tools, only using containers so the husky hooks weren't executing.  Fixed the formatting and squashed the commits.

kris7t:
This was implemented I think because Google had some validation for the User-Agent string (they really don't like people log in to Google with Electron apps directly and they'd prefer we used OAuth) that this forwarding hack worked around. In the meantime, Google probably updated its validation, and now it seems Electron is working differently, so we can't use the same hack for handling forwarding...

Note that using OAuth (as required by Google) is impossible here, because that would only give Ferdium access to the user's Google data, but it wouldn't give the browser session inside Ferdium a valid Google session cookie (this separation of privileges is the point of OAuth). SSO with Google requires a valid Google session cookie inside the browser container.

src/models/UserAgent.ts
mbentley: I believe this has been addressed; if I am mistaken, let me know but I will go ahead and resolve this.

mbentley:
I just wanted to check back in for any feedback on anything that potentially should be done differently or if this is good as is. I was reminded of this annoyance today, having to manually execute the workaround for SSO 😉

SpecialAro (approved):
I haven't tested this but LGTM!

Alphrag:
Just chipping in on that: I am afraid the removal of the `chromelessUserAgent` variable might have some unexpected side-effects. The reason why I am saying this is because the `userAgent(..)` method is called in other parts of the code, and in such cases, it will never receive the `userAgentWithoutChromeVersion` value with the proposed changes. So my question is then: are we sure that the `will-navigate` event will be fired every time we actually care about changing this user agent, and will it impact any other calls throughout the app?

I haven't been able to test it fully since I do not have access to the type of accounts having the problem, but using this fix hasn't affected anything on my side anyway. So if my remark above if of no concern, then I'm happy for this to be merged.

mbentley:
Sorry for the AI generated analysis - someone will want to double check these points but this is what Claude found:

"The userAgent(..) method is called in other parts of the code, and in such cases, it will never receive the userAgentWithoutChromeVersion value"

Places the `userAgent` getter is called:

| File | Line | Code | Purpose |
|------|------|------|---------|
| `ServiceWebview.tsx` | 145 | `useragent={service.userAgent}` | Initial webview HTML attribute |
| `Service.ts` | 397 | `return this.userAgentModel.userAgent` | Passthrough getter |
| `TodosScreen.tsx` | 40 | `userAgent={todosStore.userAgent}` | Initial todos webview attribute |
| `todos/store.ts` | 76 | `return this.userAgentModel.userAgent` | Passthrough getter |

1. These look to be initial webview creation. The `useragent` attribute on `<ElectronWebView>` sets the user agent *when the webview is first mounted*. Google SSO happens during *runtime navigation*, not at mount time.

2. The old code also set `this.webview.userAgent` directly. The getter was only used to compute the value:
   ```typescript
   // OLD CODE
   this.chromelessUserAgent = true;
   this.webview.userAgent = this.userAgent;  // <-- Direct assignment to webview
   ```

3. The fix does the exact same thing, just more directly:
   ```typescript
   // NEW CODE
   this.webview.userAgent =
     this.serviceUserAgentPref || this.userAgentWithoutChromeVersion;
   ```

4. The getter never "pushed" the value anywhere. MobX reactivity doesn't automatically update the webview's user agent property when `chromelessUserAgent` changes. The old code explicitly called `this.webview.userAgent = this.userAgent` after setting the boolean—the change just removes the intermediate state variable.

---

"Are we sure that the will-navigate event will be fired every time we actually care about changing this user agent?"

### The old code relied on the exact same events

```typescript
// OLD CODE
_addWebviewEvents(webview: ElectronWebView): void {
  this._willNavigateListener = event => this._handleNavigate(event.url, true);
  webview.addEventListener('will-navigate', this._willNavigateListener);

  this._didNavigateListener = event => this._handleNavigate(event.url);
  webview.addEventListener('did-navigate', this._didNavigateListener);
}
```

### My code uses the exact same events

```typescript
// NEW CODE
this._willNavigateListener = event => this._handleNavigate(event.url);
webview.addEventListener('will-navigate', this._willNavigateListener);

this._didNavigateListener = event => this._handleNavigate(event.url);
webview.addEventListener('did-navigate', this._didNavigateListener);
```

The Google SSO flow is a series of redirects. Every redirect triggers `will-navigate` before the request is sent, allowing the user agent to be set in time. The `did-navigate` event acts as a backup to catch any missed navigations.

---

Summary

| Aspect | Old Code | New Code | Change? |
|--------|----------|----------|---------|
| Events used | `will-navigate`, `did-navigate` | `will-navigate`, `did-navigate` | No change |
| How user agent is applied | `this.webview.userAgent = this.userAgent` | `this.webview.userAgent = <value>` | Same mechanism |
| Where getter is used | Initial webview creation only | Initial webview creation only | No functional change |

The concern stems from thinking the getter was doing something special. It wasn't—the getter was just a computed property. The actual user agent change always happened via direct assignment to `this.webview.userAgent`, which this fix preserves. The only behavioral change is removing the `loadURL()` call on `will-navigate`, which was causing the original bug (it cancelled POST requests needed for SAML authentication).

Randomness0x:
I just tested this commit and it works for me for using Google workspace in combination with keycloak. 
For Claude which also didn't work redirects me to an external browser, so this still fails. Even I set the option to open the link within Ferdium itself. But luckily Claude has a workaround by using a code sent to my e-mail.

Edit: Now running for a week and still no issues

tildebyte:
What's required to merge this?

This SSO/SAML/etc. login issue is the ONLY thing which prevents me from completely endorsing Ferdium to friends/family/coworkers/etc.

Alphrag (approved):

Alphrag:
Sorry, I thought I had merged that last time, and since I haven't had any time to spend on Ferdium lately, I was not reminded of unfinished business.
I've now merged it, which will release the nightly containing the fix tomorrow, and I will make a stable version for it to be more broadly available at the end of the week if there are no problems with that.

Thanks for chasing that!

@all-contributors please add @mbentley for code

zkeying:
TYVM for this fix!  I no longer have to perform console surgery just to login to my Google Apps!

paoloantinori:
I was waiting for this feature, thanks for working on it.
It works only partially for me, and I wonder if there are workaround I can put in place pure client side or with configuration.

When I add a new instance of Gmail, the url in the tab is https://workspace.google.com/intl/en-US/gmail/

If I click on the Sign In button, the navigation jumps out of Ferdium to my Desktop browser (Firefox on Fedora 43), even if I have configured the service to open links internally.

This prevents me from authenticating.

Interestingly, I don't have the same issue on Gemini, that allows the SSO authentication to stay in Ferdium.


Any idea?
Vesion: 7.1.2-nightly.5
System: Linux, Fedora 43 with Gnome + Firefox

paoloantinori:
Actually found the workaround, it was dead simple: just copy back the url that open in the external browser back in Ferdium 😅

Still, it would be cool if Ferdium could fully intercept links and open them directly

tildebyte:
Huzzah, fully fixed :tada:

ilopmar:
This is not working for me. After login, either with the old version doing the workaround or with the latest nightly version, there's a new request and Ferdium gets stuck at it:
```
https://accounts.google.com/o/saml2/continue?idpid=xxxxxx&SAMLRequest=xxxxxx&RelayState=vxxxxxxx.login..exxxxx&omethod=GET
```

Any help?

mbentley:
Well funny enough, there was a SSO provider change at my company and it stopped working for me again.  I was able to get it working again with the change I pushed to https://github.com/mbentley/ferdium-app/tree/another-google-fix.  Not sure if you have the ability to build that to test.

*edit: if needed, i could share a build but obviously trusting a stranger's code/build is a risk.

ilopmar:
@mbentley Thanks! How should I test your changes? Do I need to build Ferdium locally with your last commit (https://github.com/ferdium/ferdium-app/compare/develop...mbentley:ferdium-app:another-google-fix)?

mbentley:
> @mbentley Thanks! How should I test your changes? Do I need to build Ferdium locally with your last commit ([develop...mbentley:ferdium-app:another-google-fix](https://github.com/ferdium/ferdium-app/compare/develop...mbentley:ferdium-app:another-google-fix))?

Yes, exactly.

ilopmar:
@mbentley I've built Ferdium locally with you patch and installed the .deb. It doesn't get stuck anymore in that url, but after doing the SAML SSO it redirects me to `https://xxxxx.slack.com/?sso_failed=1` and it doesn't log me in.

Any hint?

mbentley:
So it's Slack where you're trying to login to and not Google or is that just some odd SSO workflow to a Google service?  Just confirming - I may not be able to test that as I don't have Slack w/SSO integration to test against.

ilopmar:
I'm trying to login into Slack but my company uses SAML. With I click on the "Sign in with SAML" I'm redirected to Google to enter my email. Then it redirects me to Microsoft (don't ask why they use both), I enter my email, password and OTP and I'm redirected back to Slack. If I do that in the browser or in Rambox (I'm using it as a workaround but I do prefer Ferdium) it works. But inside Ferdium, it doesn't work :(

Thanks for all your help.

mbentley:
If we have too much more troubleshooting, we might want to move it to an issue or maybe a discussion just in case this goes much further to keep the PR cleaner (not sure how the maintainers would prefer things exactly) but anyway...

Unfortunately troubleshooting further is going to need some detailed information that may be a bit much to share so I understand if that's a concern and you'd be more than welcome to send it to me directly (my personal email is listed on my [github profile page](https://github.com/mbentley/)).

1. Open up Ferdium and select your service before you attempt to login
2. View > Toggle Service Developer Tools (make sure you get the right one)
3. Open the Network tab
4. You may need to check the preserve log box; probably safer to do this to make sure we capture the whole auth process
5. Go through the auth process until it fails
6. Then export a HAR of the session (the box should say sanitized so it will remove any credentials, tokens, cookie values, etc)

Here's a screen shot of the export because it took me a sec to find it:
screenshot 2026-02-26 at 5 59 09 AM

Then it would be useful to get some additional info that I'd probably manually need from you:

- The final POST to slack.com/sso/saml (or whatever Slack's ACS endpoint is) — does it have blocked > 0 or send = 0? (Unlikely given sso_failed=1 but worth confirming)
- What cookies are present on that POST request? (click on the line item which will open a new pane - go to the Cookies tab on that new pane)
  - Because the HAR is sanitized, I'd need you to pull this info manually
  - You should be able to click within the Cookies pane and do a ctrl/cmd + A to select all and copy them but you will want to manually sanitize the values
- The response from Slack's ACS (does it redirect to ?sso_failed=1 or is that a different step?)

PR #2334, opened 2026-02-08
remove loading overlay

rodrigocabraln:
#### Pre-flight Checklist

#### Description of Change

Remove the fullscreen `WebviewLoader` modal ("Loading {service}") that appears when a service is loading for the first time. This modal blocks the entire UI and almost never dismisses after the service has finished loading due to a race condition between Electron webview events (`did-start-loading`, `did-stop-loading`, `did-frame-finish-load`).

The fix removes the `WebviewLoader` component from `ServiceView.tsx` and simplifies the `TopBarProgress` indicator to show on all page loads (not just after the first load), providing a non-blocking visual loading indicator instead.

**Changes:**
- Removed the `WebviewLoader` rendering block and its import from `ServiceView.tsx`
- Removed the `!service.isFirstLoad` condition from `TopBarProgress` so it works on all loads

#### Motivation and Context

The `WebviewLoader` fullscreen overlay depends on `service.isLoading && service.isFirstLoad` to dismiss. However, Electron's `did-start-loading` event fires multiple times (for sub-resources), resetting `isLoading = true` each time, while `did-frame-finish-load` (which sets `isFirstLoad = false`) doesn't always fire reliably for all frames. This race condition causes the modal to stay visible indefinitely, forcing users to manually reload or switch services.

The `TopBarProgress` bar already exists as an alternative loading indicator and does not block user interaction.

#### Checklist

- [ ] The changes respect the code style of the project (`pnpm prepare-code`)
- [ ] `pnpm test` passes

#### Release Notes

fix: Remove blocking "Loading {service}" fullscreen modal that would get stuck indefinitely due to webview event race condition

### DIFF
--- a/src/components/services/content/ServiceView.tsx
+++ b/src/components/services/content/ServiceView.tsx
@@ -14,7 +14,6 @@ import type ServiceModel from '../../../models/Service';
 import type { RealStores } from '../../../stores';
 import MediaSource from '../../MediaSource';
 import StatusBarTargetUrl from '../../ui/StatusBarTargetUrl';
-import WebviewLoader from '../../ui/WebviewLoader';
 import ServiceDisabled from './ServiceDisabled';
 import ServiceWebview from './ServiceWebview';
 import WebviewCrashHandler from './WebviewCrashHandler';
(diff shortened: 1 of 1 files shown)


SpecialAro (changes requested):
Hello!

Instead of removing the ServiceView entirely, why not try and fix the "stuck" loading? Removing a "feature" instead of fixing a bug with said feature does not seem to me as the way we should pursue.

Thank you in advance!

tbaumann:
wow. 

I see zero benefit in fixing that "feature".

The overlay might make sense for hibernated or restricted services. But loading is 100% not a condition for which it is necessary to render the page unusable.

IMO that should 100% go in the :wastebasket: 

TBH right now I prefer a degraded experience over a non functioning one. But removing the entire overlay is for sure only a workaround

stackcoder:
The issue also affects me quite often. I can’t tell exactly how to reproduce it, but it tends to hang after a longer period of standby. When it occurs, it’s quite annoying, so I would like to steer the discussion back toward a constructive exchange of views.

Another argument is that some web apps, e.g. Outlook, already provide their own loader. So what we end up with is a loader-over-loader situation, which is something no one likes to have. Having a loader could be a per-app setting. The pragmatic and fail-safe approach would be to remove it entirely.

PR #2396, opened 2026-04-23
fix: removing will-navigate UA switch that cancelled SAML POST navigations

mbentley:
#### Pre-flight Checklist

#### Description of Change

Setting webview.userAgent in the will-navigate handler calls SetUserAgentOverride() in Electron while a navigation is pending, which cancels the navigation. This broke any cross-origin form POST to accounts.google.com, including SAML ACS endpoints such as /o/saml2/continue and /samlrp/.../acs.

Fix by removing the will-navigate listener entirely and updating the user agent only in did-navigate, after navigation completes. The chromeless UA behaviour for Google accounts is preserved; it now takes effect on subsequent requests from the loaded page rather than on the navigation itself.

#### Motivation and Context

PR #2261 didn't address all edge cases seen in #1973

When logging in with my a new to me SSO provider, login works fine but at the end, it fails just returning a blank screen with the following sort of message:

```
No flow state available TransactionId: prd-4.0.1+1015010-ACC_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

The `will-navigate` handler was calling `_handleNavigate` for every navigation on every service, not just Google ones. For any non-Google URL, `_handleNavigate` would call `this.webview.userAgent = this.serviceUserAgentPref || this.defaultUserAgent`, setting it to whatever the current UA already was. But even setting the UA to the same value calls `SetUserAgentOverride()` in Electron, which cancels any pending navigation.

This means the will-navigate bug was silently breaking any cross-origin form POST on any service — Okta SAML, ADFS, Microsoft SSO, payment gateways, etc. The fix benefits all of them, not just Google.

Fix 1 (PR #2261): Removed loadURL(url) from will-navigate because it was aborting SAML POST requests. The chromelessUserAgent flag was also removed at the same time, replaced with unconditional webview.userAgent = on every will-navigate. This stripped the accidental protection: now UA assignment fires on every single navigation, including mid-SAML-flow.

Fix 2 (this PR): Removed will-navigate entirely. Now UA is only set in did-navigate, which fires after the navigation is fully committed at that point IsLoadingIncludingInnerFrameTrees() is false, so SetUserAgentOverride can't trigger a reload. I believe this is the correct and permanent fix.

#### Checklist

- [X] `pnpm test` passes

#### Release Notes

removing will-navigate UA switch that cancelled SAML POST navigations to fix additional issues #2261 aimed to fix

### DIFF
--- a/src/models/UserAgent.ts
+++ b/src/models/UserAgent.ts
@@ -7,9 +7,6 @@ const debug = require('../preload-safe-debug')('Ferdium:UserAgent');
 
 export default class UserAgent {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  _willNavigateListener = (_event: any): void => {};
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _didNavigateListener = (_event: any): void => {};
 
   @observable.ref webview: ElectronWebView = null;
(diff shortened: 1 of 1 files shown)


SpecialAro (approved):
I have not tested this but this makes sense. Thank you!

SpecialAro:
@all-contributors please add @mbentley for code

PR #2411, opened 2026-05-04
Updated Portuguese (Brazil) translation

xgrind:
I've updated the translation.


SpecialAro:
Thanks a lot for the work you've put into updating the Brazilian Portuguese translations. The file itself looks structurally sound from what I checked, and the keys/placeholders appear to be aligned with the English source.

The main concern here is the Crowdin integration rather than the translation work itself.

`en-US.json` is our Crowdin source, while `pt-BR.json` is generated/exported from Crowdin. Because of that, merging a large direct change to `pt-BR.json` risks creating two different sources of truth. Depending on the Crowdin sync settings, these changes could either be overwritten on a future Crowdin export or potentially overwrite existing translation state in Crowdin.

Since this PR changes a significant portion of the locale file, I'd prefer that we keep Crowdin as the authoritative source for these translations.

Suggested next steps:

1. Import these Portuguese translations into the `pt-BR` language in [Crowdin](https://crowdin.com/project/ferdium-app) rather than merging the locale file directly. Let me know when you ask for access to the project so I can grant you.
2. Review the imported strings in Crowdin, especially where existing translations may already have been updated since this PR was created.
3. Let the normal Crowdin sync/export workflow generate the updated `pt-BR.json` and bring it back into the repository.
4. Once the Crowdin-generated changes are available, we can compare them with this PR and make sure none of your translation work was lost.

I also noticed at least one translation where some meaning from the English source appears to have been dropped (`settings.recipes.nothingFound`, where the suggestion to upgrade Ferdium is missing), so having the translations reviewed through Crowdin would help catch cases like that as well.

I'm closing this here mainly because of the localization workflow, not because the contribution itself isn't useful. Thanks again for taking the time to improve the Portuguese translation - I'd like to make sure we preserve that work without causing problems with the Crowdin sync later.

PR #2451, opened 2026-07-11
feat: update user agent with dynamic Ferdium version

### LINKED ISSUE #2439: Google Calendar redirect loop
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that Ferdium Recipes adheres to.
- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22recipe+%3Abookmark_tabs%3A%22+label%3A%22bug+%3Abeetle%3A%22+) for a bug report that matches the one I want to file, without success.

### Ferdium Version

7.1.2

### Recipe Name

google-calendar

### Recipe Version

2.4.5

### Last Known working version of the recipe

_No response_

### Steps to reproduce

1 add service google-calendar
2 go to the new servis 
3 the page is constantly refresh

### Expected Behavior

on the step 3 we should have to authentification google page

### Actual Behavior

on the step 3 the page is constantly refresh

### Screenshots

_No response_

### Additional Information

I found a fix

Mralexandresys:
#### Pre-flight Checklist

#### Description of Change
Updates the Ferdium user agent to dynamically include version information. The Ferdium version is now automatically retrieved from `package.json` instead of being hardcoded, and the Electron version is included in the user agent string.

This fixes the Google Calendar constant refresh issue caused by an outdated user agent string.

#### Motivation and Context
The user agent was previously missing the Ferdium and Electron version signatures and was outdated, causing service compatibility issues. This was particularly affecting Google Calendar which was constantly refreshing instead of showing the authentication page.

Fixes #2439

#### Checklist
- [x] The changes respect the code style of the project
- [x] I tested my changes locally

#### Release Notes
User agent now includes Ferdium version and automatically updates with each release. Fixes constant page refresh on Google Calendar and other services with strict user agent validation.

### DIFF
--- a/src/helpers/userAgent-helpers.ts
+++ b/src/helpers/userAgent-helpers.ts
@@ -2,12 +2,15 @@ import { cpus } from 'node:os';
 import macosVersion from 'macos-version';
 import {
   chromeVersion,
+  electronVersion,
   is64Bit,
   isMac,
   isWindows,
   osArch,
   osRelease,
 } from '../environment';
+// @ts-expect-error Cannot find module '../package.json' or its corresponding type declarations.
+import { version as ferdiumVersion } from '../package.json';
 
 const macOS = () => {
   const version = macosVersion() ?? '';
@@ -42,5 +45,5 @@ export default function userAgent() {
     platformString = linux();
   }
 
-  return `Mozilla/5.0 (${platformString}) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/${chromeVersion} Safari/537.36`;
+  return `Mozilla/5.0 (${platformString}) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/${chromeVersion} Safari/537.36 Ferdium/${ferdiumVersion} (Electron ${electronVersion})`;
 }


SpecialAro:
Hi! Thank you @Mralexandresys for your work on this.

I think this change may interfere with Ferdium’s existing Google user-agent workaround.

The relevant files are:

* `src/helpers/userAgent-helpers.ts` - this is the file modified by this PR. It generates Ferdium’s default UA. This PR changes it globally from roughly:

  `Chrome/<version> Safari/537.36`

  to:

  `Chrome/<version> Safari/537.36 Ferdium/<version> (Electron <version>)`

* `src/models/UserAgent.ts` - this contains the existing Google-specific workaround. It imports `defaultUserAgent` from `../helpers/userAgent-helpers`, and `userAgentWithoutChromeVersion` only does:

  `Chrome/<version>` → `Chrome`

  Therefore the new `Ferdium/... (Electron ...)` suffix added by this PR is **not removed**.

`src/models/UserAgent.ts` also applies the Google workaround in `_handleNavigate()` only when the URL starts with `https://accounts.google.com`, and that handler is attached to `did-navigate`.

This means the order is effectively:

`src/helpers/userAgent-helpers.ts`
→ creates `Chrome/<version> ... Ferdium/... (Electron ...)`
→ initial navigation to Google uses this UA
→ `src/models/UserAgent.ts::_handleNavigate()` runs after `did-navigate`
→ removes only the Chrome version
→ resulting UA is still `Chrome ... Ferdium/... (Electron ...)`

So Google can see the Ferdium/Electron identifiers both on the initial authentication navigation and on subsequent requests.

Could you please adjust this PR so the Google authentication workaround is preserved?

I would suggest **not appending the Ferdium/Electron identifiers directly in `src/helpers/userAgent-helpers.ts`**, since that function is also the source used by the Google workaround.

A safer implementation would be to keep the existing Chrome-compatible default UA unchanged there and add the Ferdium/Electron identifiers conditionally for non-Google requests.

Ideally, the choice should be made before the request is sent, based on the destination URL, e.g. in the request/header handling layer:

* `accounts.google.com` / Google authentication → keep the existing clean Chrome-compatible UA
* other destinations → append `Ferdium/<version> (Electron <version>)`

Doing this only inside the existing `did-navigate` handling in `src/models/UserAgent.ts` would not fully solve the problem, because `did-navigate` runs after the initial Google navigation has already completed.

Mralexandresys:
Thanks for the detailed review, I hadn't seen the impact on Google sign-in.

Looking into it, Google isn't the only one affected:

- `recipes/whatsapp/index.js` already strips `(Ferdium|Electron)/...` from the UA, in `overrideUserAgent()` and `modifyRequestHeaders()`. [whatsapp/index.js](https://github.com/ferdium/ferdium-recipes/blob/main/recipes/whatsapp/index.js) 
- Outlook and Teams serve a degraded version to browsers they don't recognise.
- Cloudflare and reCAPTCHA react badly to an unusual UA.

So enabling it by default looks like a mistake to me, even with the Google fix.

I'd rather turn it into an option that's off by default: a toggle in Settings → Advanced, a per-service override (inherit / yes / no) like `spellcheckerLanguage`, the string still generated dynamically so it doesn't go stale, and the Google workaround moved into `onBeforeSendHeaders` filtered on `accounts.google.com` as you suggested.

Would you rather I rework the PR along those lines, close it and keep the current behaviour, or keep it as a default with just the Google fix?

SpecialAro:
Thanks for taking the time to dig into this further, and especially for checking the behavior across other services. I appreciate the work you've put into the PR and the thoughtful follow-up.

Given what we've uncovered, I would lean toward closing/replacing the current approach and keeping the existing default UA behavior, while handling the Google authentication issue separately.

My main concern is that adding the Ferdium/Electron identifiers globally changes this from a targeted compatibility fix into a broader UA policy decision, with potential side effects across services. I also think introducing the global toggle plus per-service inheritance/overrides would add quite a bit of complexity to this PR.

For Google specifically, moving the workaround to the request-header layer sounds like the right direction, since it can handle the initial authentication request as well. I'd prefer to solve that as a focused fix first.

If exposing Ferdium/Electron in the UA is something we want as a feature, I think it would be better considered separately as an explicit opt-in feature, with the interaction between global UA settings, per-service overrides, and recipe-level UA handling clearly defined.

Thanks again for investigating this and adapting the proposal based on the feedback.

Issue #1608, opened 2024-03-11
Sometimes apps get stuck in the "Loading" status 


MurzNN:
### Avoid duplicates

- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

6.7.1

### What Operating System are you using?

Ubuntu

### Operating System Version

Kubuntu 23.10

### What arch are you using?

x64

### Last Known Working Ferdium version

It happens on several previous versions too.

### Expected Behavior

Apps should load well, or provide an ability to reload it and get rid of the popover spinner.

### Actual Behavior

Sometimes a specific app (Gmail, Slack, Telegram, MS Outlook) can get stuck in the "Loading [appname]" status like this:
![image](https://github.com/ferdium/ferdium-app/assets/336662/ffc035d9-3f59-4732-bde8-6785d7df169e)

Usually, it happens on the Ferdium startup for some (not all) services, but sometimes - it happens during work at random times.

Clicking the "Reload" menu item on the service right-click menu doesn't resolve the problem - the page reloads, but still with the constant "Loading [appname]" popover with no ability to get rid of it, or click on the webpage elements.

The same with the menu item "View » Reload Service".

The only workaround with clicking "Disable service" and then "Enable service" helps with resolving this problem.

### Steps to reproduce

1. Load Ferdium
2. See the infinite  "Loading [appname]" popup spinner and locked service.

### Debug link

https://debug.ferdium.org/200d7dcf-f0ad-463c-89e5-9ce5cbe2cb7f

### Additional information

Version: 6.7.1
Server: Ferdium Server
Electron: 29.1.1
Chrome: 122.0.6261.111
Node.js: 20.9.0
Platform: Linux 6.5
Arch: x64
Build date: Sun Mar 10 2024 08:03:46 GMT+0400 (Armenia Standard Time)
Git SHA: f1f02563
Git branch: release

kertzi:
Hello, I'm hitting this also. 
It's wierd, I added two Google calendar, one works, another just "loading" and "loading". 
Same with Teams. 
Just downloaded latest version from https://ferdium.org/download 
[...]

PR #2460, opened 2026-07-28
fix: service is permanently stuck on loading screen

eXistenZNL:
#### Pre-flight Checklist

#### Description of Change
Fix for the loading screen that is sometimes stuck.

This happens because `getWebContentsId()` throws an Error if the <webview> isn't fully attached yet (a known Electron timing issue, electron/electron#31918), and when it throws the Error, the function bails out before registering a single webview event listener. So nothing is ever left in the app that could clear the loading state for that service.

#### Motivation and Context
Me, along with many others, experience this bug and when it happens, the only solution is to reload the service (helps sometimes) or restart Ferdium as a whole (helps often). For me personally, with 9 services, this happens 90% of the time to at least 1 service when I start Ferdium in the morning.

This was also how I was able to verify that this fixes the bug 100% of the time. I started a patched Ferdium 10 times in a row which is 90 services, and none of them failed. The original Ferdium has at least 1 or 2 services stuck on every boot.

This issue is described in #1608, and an attempt to completely remove the loading screen was done in #2334 however @SpecialAro wanted to keep the Loading overlay as-is and instead asked to fix the root cause of this issue so that's why I dove into the issue.

#### Checklist

- [x] `pnpm test` passes

#### Release Notes

Fix for the loading screen that is sometimes stuck

### DIFF
--- a/src/models/Service.ts
+++ b/src/models/Service.ts
@@ -414,9 +414,35 @@ export default class Service {
   }
 
   initializeWebViewEvents({ handleIPCMessage, openWindow, stores }): void {
-    const webviewWebContents = webContents.fromId(
-      this.webview.getWebContentsId(),
-    );
+    let webviewWebContents;
+    try {
+      webviewWebContents = webContents.fromId(this.webview.getWebContentsId());
+    } catch (error) {
+      // The <webview> can still be mid-attach here even after the
+      // setTimeout(0) workaround in ServiceWebview's onDidAttach (see
+      // https://github.com/electron/electron/issues/31918). getWebContentsId
+      // throws in that case instead of returning a usable id. Rather than
+      // assuming a fixed delay is enough (and silently never initializing
+      // the service, leaving it stuck in its default "loading" state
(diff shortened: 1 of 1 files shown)


SpecialAro:
Thanks @eXistenZNL for working on this.

I added a small lifecycle guard around the `dom-ready` retry.

The retry now captures the webview that originally failed initialization and verifies that it is still the active webview before calling `initializeWebViewEvents()` again. This prevents a detached or replaced webview from initializing listeners on a newer instance.

The change also:
- Returns safely when the webview has already been detached.
- Uses the captured webview when retrieving and storing its WebContents.
- Preserves the existing one-time `dom-ready` retry behavior.
- Avoids duplicate listener registration in the normal retry path.

SpecialAro (approved):

eXistenZNL:
Sure, makes sense. I tested the changes that you pushed on top and it still works great.

Feel free to merge!

SpecialAro:
@all-contributors please add @eXistenZNL for code

Issue #2295, opened 2025-11-19
Google/Android Messages css not injecting


jed84:
### Avoid duplicates

- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

7.1.1

### What Operating System are you using?

Windows

### Operating System Version

Windows 25H2

### What arch are you using?

x64

### Expected Behavior

Injected custom css works in Google Messages

### Actual Behavior

Injected custom css does not work in Google/Android Messages

### Steps to reproduce

Use the user.css file in attachment

### Additional information

[user.css](https://github.com/user-attachments/files/23616391/user.css)

jed84:
This can be fixed for Android Messages and other Google services (such as Google Calendar) by modifying the code at the top of the 'webvew.js' file so it starts like this:

```
const _interopRequireDefault = (obj) => obj && obj.__esModule ? obj : { default: obj };

const _path = _interopRequireDefault(require('path'));

  // Import your addons.js CSS content
const { cssContent } = require('./addons.js');

module.exports = Ferdium => {
```

and ends like this:

```
      document.querySelector(`#${cssId}`)?.remove();
    }
  });

  // Inject CSS from addons.js
  const style = document.createElement('style');
  style.type = 'text/css';
  style.textContent = cssContent;
  document.head.appendChild(style);
};
```

Then we need to create a separate file 'addons.js', where we can add CSS changes.

An example of 'addons.js' file:
```

const css = `

/* Top Navigation: Logo hide */
h1.logo.no-margin
{ display: none !important; }

/* Sidebar: Battery info hide */
.banner-container
{ display: none !important; }


`;

function injectCustomCss() {
  const style = document.createElement('style');
  style.id = 'custom-css-fixes';
  style.type = 'text/css';
  style.appendChild(document.createTextNode(css));
  document.head.appendChild(style);
}

// Run immediately
injectCustomCss();

module.exports = { injectCustomCss };
```

In this way, we can prevent CSS settings from being overwritten by Google.

I wonder if this can be included in future updates of Google services in Ferdium?

Please review/consider this as a starting point to provide a fix.

Issue #2288, opened 2025-11-12
Messenger css not injecting


playagain96:
### Avoid duplicates

- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

7.1.1

### What Operating System are you using?

Windows

### Operating System Version

Windows 11 25H2

### What arch are you using?

x64

### Expected Behavior

Im trying to inject custom css in messenger downloaded from Userstyles Archive, which works on regular browser with Stylus extension
This is the css:
[user.css](https://github.com/user-attachments/files/23504344/user.css)

### Actual Behavior

In Ferdium by using the user.css file - it doesnt work

### Steps to reproduce

Use the user.css file in attachment

Issue #1607, opened 2024-03-08
Google User.css NOT LOADER


Nickyg001:
### Avoid duplicates

- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

6.7.0

### What Operating System are you using?

Windows

### Operating System Version

windows 10

### What arch are you using?

x64

### Expected Behavior

until a few days ago I used the css customization applied to the "user.css" file to customize and enlarge the fonts within GMAIL and everything worked very well. It hasn't worked for 2-3 days now and I see small texts by default.

1. What happened and how do I solve it?
2. Is it not possible to make the inserted apps adapt to the monitor resolution? with 4k everything becomes small and difficult to read.
3. In Chrome using Gmail I have the possibility to scale the size of the UI with CTRL(+), wouldn't it be possible to do the same thing in Ferdium too?

Thank you

### Actual Behavior

view small font (default)

### Steps to reproduce

.

SpecialAro:
@Nickyg001 have you look to the View > Zoom in and Zoom out options? Is that what you need?

Nickyg001:
> @Nickyg001 have you look to the View > Zoom in and Zoom out options? Is that what you need?

I would like the parameters contained in the file user.css to be read
but it doesn't seem like he's doing it.

alternatively I can't find this in the menu, how do I do it?   View > Zoom in and Zoom out options?

mmoayyed:
I am seeing the same problem on MacOS. `~/Library/Application Support/Ferdium/recipes/gmail/user.css` is not loaded.

```
Version: 6.7.1
Server: Ferdium Server
Electron: 29.1.1
Chrome: 122.0.6261.111
Node.js: 20.9.0
Platform: macOS Unknown
Arch: arm64
Build date: Sun Mar 10 2024 06:30:21 GMT+0400 (Gulf Standard Time)
Git SHA: f1f02563
Git branch: release
```

EzioTheDeadPoet:
This issues also impacts the the following services:

- Discord
- MS Teams

```
Version: 6.7.1
Server: No Server
Electron: 29.1.1
[...]

Issue #1102, opened 2023-04-09
user.css and user.js is not loading on vk.com


miztizm:
### Avoid duplicates

- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

6.2.6

### What Operating System are you using?

Windows

### Operating System Version

windows 11

### What arch are you using?

x64

### Last Known Working Ferdium version

ferdi

### Expected Behavior

Hello, I made a user.css and user.js in the VK recipe, and added several accounts from VK. But when I tried to open a app, some times the user.css seems to be working, sometimes it not loaded at all—but then when reload of service is triggered, all traces of the user cosmetic code would simply vanish away without fail as well afterwards; unfortunately. I switched from Ferdi and this all works just fine before. 

### Actual Behavior

user.css and user.js dosent work

### Steps to reproduce

1. open ferdium
2. one load it works another load it dosent work.
3. user.css and user.js dosent work

### Screenshots

![image](https://user-images.githubusercontent.com/617020/230795262-5be784f5-5010-4ed5-82df-cf35528e20f2.png)
![image](https://user-images.githubusercontent.com/617020/230795412-eb6a7e7b-a07f-4f98-bb44-2c5645990eb5.png)

### Additional information

user.js

```
module.exports = (config, Ferdium) => {
  // Write your scripts here
    alert('ok');
};

```

user.css
`
body { background: #222222!important; }`

vraravam:
What is the "VK" recipe? Please do not use abbreviations that are not understood.

miztizm:
> What is the "VK" recipe? Please do not use abbreviations that are not understood.

Its a social network VK.COM @vraravam 

![image](https://user-images.githubusercontent.com/617020/230849526-7da8b576-ce98-4e2a-8408-2077a8c96db8.png)

i just created 2 different accounts with the user.css and user.js on messenger service recipe and it works fine, but not with VK.com

How can i debug this?

warshell:
How can i debug this?

miztizm:
Hello , any news or suggestions to debug this ? thanks

PR #2463, opened 2026-08-05
fix: fill injected <style> elements via textContent, not innerHTML

tfluehmann:
#### Pre-flight Checklist

#### Description of Change

Every place we fill an injected `<style>` element used `innerHTML`, which is a [Trusted Types](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API) injection sink. On a service whose document sends `require-trusted-types-for 'script'`, that assignment throws and the stylesheet is **silently never applied**.

`textContent` is not a sink, and for a `<style>` element it is an equivalent way to set the stylesheet text — so this is a drop-in swap, four call sites:

| file | what silently fails today |
|---|---|
| `src/webview/darkmode.ts` | `darkmode.css` never applied |
| `src/webview/lib/RecipeWebview.ts` | recipe `Ferdium.injectCSS()` never applied |
| `src/webview/recipe.ts` | `user.css` never applied |
| `src/webview/contextMenuBuilder.ts` | translator popup renders unstyled |

The `recipe.ts` one is the worst of the four. The throw escapes `loadUserFiles()` before it reaches the `user.js` block further down the same function, so **a single `user.css` file takes `user.js` down with it** — which is precisely the traceback in #1086.

`src/features/appearance/index.ts` also fills a `<style>` via `innerHTML`, but I deliberately left it alone: it styles Ferdium's own renderer document, which no third-party CSP applies to.

#### Motivation and Context

#1086 reported `user.js` failing to load on Google Calendar with:

```
TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment.
    at v.loadUserFiles (recipe.js:1:4621)
```

It was closed as fixed by ferdium/ferdium-recipes#402, but that PR only routed *one recipe's own* stylesheet around the problem by hosting it on a CDN — `recipe.ts:270` was never touched, so `user.css`/`user.js` are still broken on any such service. This addresses the actual sink.

This is not hypothetical or spec-lawyering — I verified it in Electron 37 (the version this repo pins) against a page carrying that exact CSP:

```json
{
  "innerHTML":            "TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment.",
  "textContent":          "OK",
  "textContentPlusEquals": "OK",
  "appliedFontWeight":    "700",   // textContent stylesheet took effect
  "appliedColor":         "rgb(0, 0, 0)",  // innerHTML stylesheet did NOT (would be rgb(1,2,3))
  "appliedLetterSpacing": "3px"    // the `+=` form used in recipe.ts also works
}
```

Note the error string is byte-identical to the one in #1086, and `appliedColor` confirms the failure mode is a silently-lost stylesheet rather than a visible crash.

Scope note, since I'd rather under-claim: I only have hard evidence for services that actually send this CSP. `messages.google.com` sends `require-trusted-types-for 'script'` as a response header today (which matches open #2295), and Google Calendar demonstrably enforces it per #1086's traceback. There are several other open "css not injecting" reports (#1607, #1102, #1060, #956, #2288) that *may* share this cause, but I could not confirm the CSP on those origins, so I'm not claiming this fixes them.

#### Checklist

- [x] `pnpm test` passes (6 suites, 81 passed, 2 skipped)

One caveat worth flagging for reviewers: these four files are at 0% test coverage, so there is no automated test proving the behaviour change — hence the Electron probe above. Happy to add a regression test if you want one, though it would need a DOM test environment (the suite currently runs `testEnvironment: 'node'`).

Behavioural difference to be aware of: `innerHTML` decodes HTML entities while `textContent` does not, so a stylesheet literally containing something like `&gt;` would now be taken verbatim. That is the correct reading for CSS, and no bundled stylesheet relies on it.

#### Release Notes

Fixed `user.css`, `darkmode.css` and recipe stylesheets not being applied on services that enforce Trusted Types (such as Google Calendar and Google Messages), which also prevented `user.js` from loading.

### DIFF
--- a/src/webview/contextMenuBuilder.ts
+++ b/src/webview/contextMenuBuilder.ts
@@ -99,7 +99,10 @@ const translatePopup = (res, isError: boolean = false) => {
   }
 
   const style = document.createElement('style');
-  style.innerHTML = `
+  // textContent, not innerHTML: innerHTML is a Trusted Types sink, so sites
+  // sending `require-trusted-types-for 'script'` reject the assignment and the
+  // translator popup renders unstyled.
+  style.textContent = `
     .container-ferdium-translator {
       position: fixed;
       opacity: 0.9;
--- a/src/webview/darkmode.ts
+++ b/src/webview/darkmode.ts
@@ -23,7 +23,10 @@ export const injectDarkModeStyle = (recipePath: string) => {
     const data = readFileSync(darkmodeCss);
     const styles = document.createElement('style');
     styles.id = ID;
-    styles.innerHTML = data.toString();
+    // textContent, not innerHTML: innerHTML is a Trusted Types sink, so sites
+    // sending `require-trusted-types-for 'script'` reject the assignment and
+    // darkmode.css is never applied.
+    styles.textContent = data.toString();
     debug('Loaded darkmode.css from: ', darkmodeCss);
 
     document.querySelector('head')?.append(styles);
--- a/src/webview/lib/RecipeWebview.ts
+++ b/src/webview/lib/RecipeWebview.ts
@@ -131,7 +131,10 @@ class RecipeWebview {
     files.forEach(file => {
       if (pathExistsSync(file)) {
         const styles = document.createElement('style');
-        styles.innerHTML = readFileSync(file, 'utf8');
+        // textContent, not innerHTML: innerHTML is a Trusted Types sink, so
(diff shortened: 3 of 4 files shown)


SpecialAro (approved):
LGTM! Thank you @tfluehmann for your work on this.😄

SpecialAro:
@all-contributors please add @tfluehmann for code

PR #2525, opened 2026-08-27
fix: restore service event handlers in shared and recreated webviews

OskarEichler:
#### Pre-flight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md).
- [x] I agree to follow the project's [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md).

#### Description of Change

Replace the partition-path registry with a weak set of initialized WebContents.

The `login` and `before-input-event` listeners are emitted by individual WebContents, not by their session. Deduplicating by storage path skipped registration for later services in a shared sandbox (including `persist:general-session`) and for replacement webviews after hibernation.

Each WebContents now receives these handlers once, while download listeners retain their existing session-level deduplication. The weak set also avoids retaining partition-path strings indefinitely.

#### Reproduction and Verification

An isolated check creates three distinct WebContents with the same storage path and initializes each twice. Before this change, handler counts are 1/0/0; after it, both login and keyboard handler counts are 1/1/1, without duplicate registration.

Validated independently on this branch with Node 24.18.1 / pnpm 11.20.0:

- `pnpm typecheck`.
- ESLint with zero warnings, Prettier, and Biome on the changed files.
- Existing Jest suite: 14 suites passed; 119 tests passed, 2 existing skips.
- Source build via `preval-build-info-cli` and `node esbuild.mjs`.
- `git diff --check`.

The Node-only Jest run used `ELECTRON_OVERRIDE_DIST_PATH="$PWD/node_modules/electron/dist" pnpm test --runInBand --coverageReporters=text-summary` to avoid Electron's lazy binary download. No Electron GUI was launched, and native cross-platform interaction / packaged installers were not tested. No test files were added or changed; the targeted old/new comparisons were isolated runtime checks.

#### Release Notes

Restore proxy authentication and zoom shortcut handlers for services sharing a session and for recreated webviews.


SpecialAro (approved):
I’m going to merge this, as the per-`WebContents` deduplication is the correct fix for the original issue.

That said, I want everyone to be aware of a few related behaviors I found while reviewing the surrounding code.

### Known bug

**Zoom shortcut interception can interfere with service-specific shortcuts**

Once this PR is merged, the `before-input-event` handler will correctly be attached to every relevant `WebContents`.

That also means Ferdium will consistently intercept `Ctrl + +`, call `preventDefault()`, and handle zoom itself for all services.

If a service/web application uses that shortcut for its own functionality, it will no longer receive the event. Previously this could appear to work simply because the handler was missing from some `WebContents` due to the partition-level deduplication bug.

I consider this a separate bug that we should track.

### Potential issues worth investigating

**Authentication handling exists at both `WebContents` and app level**

We currently handle `login` events both in `Service.ts` and via the global `app.on('login')` handler.

Both paths call `preventDefault()`, and the global implementation also uses a single shared `authCallback`. Concurrent authentication requests could potentially interfere with each other.

This is not introduced by this PR, but the authentication flow would benefit from being reviewed and consolidated.

**`WeakSet<WebContents>` depends on stable remote object identity**

Using a `WeakSet<WebContents>` is the correct model here, and `@electron/remote` should preserve the proxy identity for a live remote object.

Still, this would be a good area for regression coverage, especially around hibernation/re-attachment and repeated initialization of the same service.

**The service-level `login` handler prevents default handling even when it does not handle the request**

The handler calls `event.preventDefault()` before checking whether it is actually a proxy/basic-auth challenge or whether credentials are available.

This appears to be pre-existing rather than caused by this PR, but it could produce unexpected authentication behavior in edge cases.

**Possible proxy Session mismatch**

`serviceProxy` configures proxies against:

```ts
session.fromPartition(`persist:service-${service.id}`)
```

while a service's actual WebContents may be using `persist:general-session` or a shared `persist:sandbox-*` partition.

That means the Session where the proxy is configured may not always be the Session the service is actually using. This also predates this PR, but it looks worth investigating separately.

None of these change my decision on this PR: the original partition-based listener deduplication is incorrect, and this PR fixes it. I’m merging, but we should track the points above as follow-up work.

SpecialAro:
Follow-up items for this PR:

* [ ] **Fix zoom shortcut interception**

  * Review the global handling of `Ctrl + +`.
  * Avoid breaking service-specific shortcuts when Ferdium handles zoom.
  * Verify behavior across Windows/Linux and different keyboard layouts.

* [ ] **Review authentication event ownership**

  * Audit the interaction between the `WebContents` `login` handler in `Service.ts` and the global `app.on('login')` handler.
  * Confirm there are no duplicate/conflicting authentication flows.

* [ ] **Fix concurrent authentication handling**

  * Replace/rework the single shared `authCallback`.
  * Ensure simultaneous authentication challenges cannot overwrite each other.

* [ ] **Review unconditional `login.preventDefault()`**

  * Only suppress Electron's default authentication handling when Ferdium is actually going to handle that challenge.
  * Check behavior when proxy credentials are missing or when the auth type is unsupported.

* [ ] **Verify proxy Session/partition consistency**

  * Confirm that `serviceProxy` configures the same Electron `Session` actually used by the service WebContents.
  * Cover `persist:general-session`, `persist:service-*`, and shared `persist:sandbox-*` cases.

* [ ] **Add lifecycle regression coverage**

  * Same `WebContents` initialized multiple times should only receive handlers once.
  * A replacement `WebContents` after hibernation/re-attachment should receive handlers.
  * Multiple WebContents sharing the same Session/partition should each receive their own handlers.

* [ ] **Verify `WeakSet<WebContents>` identity assumptions**

  * Confirm `@electron/remote` preserves stable object identity for the lifetime required by the deduplication logic.

PR #2530, opened 2026-08-27
perf: reduce blocking work when switching services

OskarEichler:
#### Pre-flight Checklist

- [x] I have read the Contributing Guidelines.
- [x] I agree to follow the Code of Conduct.

#### Description of Change

- Cache each service's page title, refresh it on navigation and `page-title-updated`, and clear it when its webview detaches.
- Update the window title in a separate reaction. Background services no longer replace the active service's window title.
- Focus a newly selected service once. Preserve current-tab clicks, window refocusing, split-view scrolling, Todos, and load-completion focus after hibernation.
- Remove the delayed load-completion title query that could run after a different service became active or the original webview detached.

#### Motivation and Context

An isolated execution of the existing ServicesStore/MobX switching path showed two synchronous `getTitle()` calls and two blur/focus cycles for one warm service switch. The explicit activation handler and the active-service reaction both performed this work. The same reaction also refocused a service when its name or dialog title changed.

With this change, the same diagnostic reports **0 title queries and 1 blur/focus cycle per warm switch**. Title-only changes cause no focus calls. Navigation still seeds the cached title with `getTitle()`; it is no longer queried during ordinary activation.

This targets Ferdium's host-side overhead, not Slack/WhatsApp rendering. Hibernation settings and memory-saving behavior are unchanged; waking an unloaded service still requires loading its page.

#### Validation

- Node 24.18.1 / pnpm 11.20.0.
- `pnpm typecheck`.
- Targeted ESLint (zero warnings), Biome, Prettier, and `git diff --check`.
- Existing Jest suite: **14 suites passed; 119 tests passed, 2 existing skips**.
- Source build: `preval-build-info-cli` followed by `node esbuild.mjs`.
- React Doctor: no findings on the changed files.
- Additional inline diagnostics using actual store/component methods and MobX, with browser APIs mocked: warm switching, same-tab clicks, window refocus, foreground/background title changes, split scrolling, hibernation wake/load focus, listener replacement/removal, detach cleanup, and Todos focus.

No new or modified test files. No packaged-app GUI run or real-world latency measurement is claimed. Native verification should exercise rapid switching, keyboard focus, split view, and hibernation on each desktop platform.

#### Checklist

- [x] My pull request is properly named.
- [x] Targeted code-style checks pass (rather than repository-wide autofixing).
- [x] `pnpm test` passes.
- [ ] Packaged application manually previewed.

#### Release Notes

Reduce redundant focus and title work when switching services, and prevent background pages from overwriting the active window title.

### DIFF
--- a/src/components/services/content/ServiceWebview.tsx
+++ b/src/components/services/content/ServiceWebview.tsx
@@ -66,14 +66,16 @@ class ServiceWebview extends Component<IProps> {
     debug('Service logged a message:', e.message);
   };
 
-  handleDidNavigate = (): void => {
+  @action handleDidNavigate = (): void => {
     if (this.props.service._webview) {
-      document.title = `Ferdium - ${this.props.service.name} ${
-        this.props.service.dialogTitle
-          ? ` - ${this.props.service.dialogTitle}`
-          : ''
-      } ${`- ${this.props.service._webview.getTitle()}`}`;
+      this.props.service.pageTitle = this.props.service._webview.getTitle();
     }
+  };
+
+  @action handlePageTitleUpdated = (
+    event: Electron.PageTitleUpdatedEvent,
+  ): void => {
+    this.props.service.pageTitle = event.title;
   };
 
   handleWebviewRef = (webview: ElectronWebView | null): void => {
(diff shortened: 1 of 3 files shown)


SpecialAro (approved):
LGTM! Thanks!

SpecialAro:
@all-contributors please add @OskarEichler for code

PR #2532, opened 2026-08-27
fix: recover failed service loads automatically and on tab activation

OskarEichler:
#### Pre-flight Checklist

- [x] I have read the Contributing Guidelines.
- [x] I agree to follow the Code of Conduct.

#### Description of Change

Recover transient **main-frame page-load failures** without requiring the user to keep pressing Cmd/Ctrl+R.

- Set `isError` when a main-frame load fails, enabling the existing error details and Reload button instead of silently showing a blank page.
- Retain the error through loading activity and error-document completion. Clear it only when a real main-frame navigation commits; reset the retry budget only after that main frame finishes loading.
- Use the existing 10-second service-maintenance tick for at most three automatic retries, with minimum delays of 10, 30 and 60 seconds after successive failures. No additional per-service timers or healthy-page IPC probes.
- Retry an eligible failed service when it is selected, including after the automatic budget is exhausted. A 10-second per-service cooldown and an in-flight guard prevent rapid switching from causing overlapping/repeated loads. Selecting a tab does not reset the automatic budget.
- Pause recovery while offline and skip detached, disabled, hibernating, loading, crashed, access-restricted, media-playing and Todos services.
- Allowlist transient Chromium network errors (including `ERR_NETWORK_CHANGED`, previously ignored). Certificate, permission, authentication, blocked-request and other unclassified failures remain visible for manual intervention. Canceled navigations and subframe failures do not trigger recovery.
- Do not automatically reload a document that already committed and subsequently reports a loading failure: it may still contain usable state or unsaved work.
- Reopen the configured service URL using a fresh GET. Do not replay failed POST bodies or authentication callback URLs.
- Ignore lifecycle events from replaced webviews and consume retry promise rejections without letting stale completions alter newer navigation state.

#### Motivation and Context

A failed service navigation can leave Chromium's blank internal error document while Ferdium's `_didFailLoad` explicitly sets `isError = false`. Neither service activation nor maintenance retries that navigation. A normal service reload can recover it, but currently requires repeated manual intervention.

This PR handles that specific failure mode. It does not claim to prevent upstream website failures or recover arbitrary JavaScript deadlocks. It does not change the separate recipe-heartbeat watchdog, renderer crash recovery, or recipe scripts.

Electron's [navigation event contract](https://www.electronjs.org/docs/latest/api/webview-tag#event-did-fail-load) and [43.3.0 navigation implementation](https://github.com/electron/electron/blob/v43.3.0/shell/browser/api/electron_api_web_contents.cc#L2356) distinguish error-document commits from successful `did-navigate` events. This change uses that distinction rather than treating any frame completion as recovery.

#### Validation

- `pnpm typecheck`
- Targeted ESLint (zero warnings), Biome and Prettier checks; `git diff --check`
- Existing Jest suite: **14 suites passed, 119 tests passed, 2 existing skips**
- Inline actual-source diagnostics with real MobX and mocked Electron events/clock: **49 scenarios passed**, covering backoff/cap/cooldown, activation and maintenance integration, offline recovery, main/subframe ordering, successful recovery, stale webviews, in-flight requests, synchronous throws/asynchronous rejections, and the safety guards.
- Source build: `pnpm exec preval-build-info-cli` followed by `node esbuild.mjs`
- React Doctor: no findings in the changed files.
- No test files added or modified. The patched packaged desktop app has not been exercised against logged-in third-party services; this is not a claim of live end-to-end verification.

Suggested desktop validation: cause a service's initial/main-frame navigation to fail with a transient network error; restore connectivity; verify bounded recovery and the Reload fallback. Repeat with automatic attempts exhausted and select the failed service. Confirm healthy tabs, subframe-only failures, certificate failures, and media-playing services are not automatically reloaded.

#### Release Notes

Retry transient service loading failures automatically and when returning to the affected tab, with bounded retries and visible error details.

### DIFF
--- a/src/models/Service.ts
+++ b/src/models/Service.ts
@@ -17,6 +17,24 @@ import UserAgent from './UserAgent';
 
 const debug = require('../preload-safe-debug')('Ferdium:Service');
 
+const LOAD_RETRY_DELAYS = [10_000, 30_000, 60_000];
+// Chromium net errors for transient network failures. Certificate, permission,
+// authentication and blocked-request errors still require user intervention.
+const RETRYABLE_LOAD_ERRORS = new Set([
+  -7, // TIMED_OUT
+  -21, // NETWORK_CHANGED
+  -100, // CONNECTION_CLOSED
+  -101, // CONNECTION_RESET
+  -102, // CONNECTION_REFUSED
+  -104, // CONNECTION_FAILED
+  -105, // NAME_NOT_RESOLVED
+  -106, // INTERNET_DISCONNECTED
+  -109, // ADDRESS_UNREACHABLE
+  -118, // CONNECTION_TIMED_OUT
+  -137, // NAME_RESOLUTION_FAILED
+  -352, // HTTP2_PING_FAILED
+]);
+
 // Global registry for active partitions
 // This is needed to prevent events of the same partition from being registered multiple times (when using custom sandboxes)
 const activePartitions = new Set<string>();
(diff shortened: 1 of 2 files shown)


SpecialAro (approved):
LGTM! Thank you for this!

PR #2412, opened 2026-05-05
fix: open OAuth popups in-app to preserve window.opener for auth callbacks

shuhikari:
#### Pre-flight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change

Make the webview-level ``setWindowOpenHandler`` in ``src/index.ts`` disposition-aware:

- ``disposition === 'new-window'`` (i.e. ``window.open()`` from JavaScript, the disposition used by OAuth popups): allow the popup as a child ``BrowserWindow`` that inherits the service ``session``, with ``parent: mainWindow``, ``fullscreenable: false`` and ``outlivesOpener: false``. This preserves ``window.opener`` so the parent page can receive the ``postMessage`` callback that completes the auth flow, and keeps the popup on the same partition so cookies are shared with the service.
- ``foreground-tab`` / ``background-tab`` (regular link clicks, ``target=_blank``, ctrl-click): unchanged — keep handing off to ``openExternalUrl`` so links open in the user's default browser.

A ``did-create-window`` listener is added so popups call ``enableWebContents`` and inherit ``webRTCIPHandlingPolicy``, mirroring the existing ``open-browser-window`` IPC path (``src/index.ts:679-680``).

#### Motivation and Context

The previous handler was unconditional: every ``window.open`` from a service was sent to the OS default browser via ``openExternalUrl``. That works for plain link clicks, but breaks any OAuth/SSO flow that opens a popup expecting ``window.opener.postMessage`` to deliver the result back to the service — the popup ended up in the system browser and the callback never reached the original webview. Users typically saw two windows opening (one external, one stale) and the service hung at the login step.

Likely fixes / closely related to:
- #2273 — Google secured log-in is not working! (Any.do and ClickUp tested)
- #1008 — [Bug]: ClickUp unable to SSO
- #1420 — Rocket.Chat oAuth window opens in browser
- #634 — Figma doesn't login with Google SSO either as the Service or as a Custom Website

This is independent from #2261 / #2396, which addressed a separate User-Agent override issue that was cancelling SAML POST navigations.

#### Screenshots

n/a

#### Checklist

- [x] My pull request is properly named
- [x] The changes respect the code style of the project (``pnpm prepare-code``)
[...]

Issue #1014, opened 2023-02-21
Can't login to notion because of browser settings


valdoryu:
### Avoid duplicates

- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22bug+%3Abeetle%3A%22) for a bug report similar to mine, in vain

### Ferdium Version

6.2.4

### What Operating System are you using?

macOS

### Operating System Version

Ventura 13.2 (a) Beta 

### What arch are you using?

arm64 (including Apple Silicon)

### Expected Behavior

When I add "Notion" service, I click the "login with google" button and it opens a popup to login with google

### Actual Behavior

When I add "Notion" service, I click the "login with google" button and it says "It looks like your browser is currently blocking popups. Please allow pop-ups to continue" making login impossible

### Steps to reproduce

Go to Notion service
Click "Login" link in the navbar
Click on "Continue with google" button

### Screenshots

<img width="1084" alt="Screenshot 2023-02-21 at 14 59 23" src="https://user-images.githubusercontent.com/7875009/220365637-11df2961-bc4d-4479-89f6-b441468eb7ee.png">

andrea-berling:
Any updates on this?

SvenAndBits:
My solution has been:
* Temporarily set the user agent to `Mozilla/5.0 (iPhone; CPU iPhone OS 16_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Mobile/15E148 Safari/604.1`. This is from an iOS device, so no popups possible. That forces notion to use the current page
* After setup/login: Revert to default user agent

It may be necessary to do this whenever the login session expires or there is some other login popup for example for connections to third party services.

andrea-berling:
@SvenAndBits Thanks man, worked like a charm 🙂

illimited-wolf:
> My solution has been:
> 
> * Temporarily set the user agent to `Mozilla/5.0 (iPhone; CPU iPhone OS 16_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Mobile/15E148 Safari/604.1`. This is from an iOS device, so no popups possible. That forces notion to use the current page
[...]

Issue #974, opened 2022-11-22
[Bug]: Slack Huddle : Can't open the screen sharing preview


ragusa87:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Ferdium Version

6.2.0

### Recipe Name

slack

### Recipe Version

b14a52d

### Steps to reproduce

1. Add a "slack" service and login
2. Join a channel and start a huddle
3. Share your own screen
4. Click on the screen preview to maximize it or on the "View Shared Screen" button (see screenshot)

### Expected Behavior

The shared screen should be opened in a new window so you can see the screen content

### Actual Behavior

Nothing happen, the new window doesn't open.

### Screenshots

![image](https://user-images.githubusercontent.com/1695207/203286563-d2b27066-24bc-4e6e-bc4c-dd0d7f29574f.png)

### Additional Information

Logs:
```
[42529:1122/105853.551585:ERROR:egl_dmabuf.cc(569)] Failed to record frame: Error creating EGLImage - EGL_NOT_INITIALIZED
[42529:1122/105853.551637:ERROR:shared_screencast_stream.cc(655)] Dropping DMA-BUF modifier: 72057594037927935 and trying to renegotiate stream parameters
(electron) The new-window event is deprecated and will be removed. Please use contents.setWindowOpenHandler() instead.

```

ragusa87:
Slack recipe on commit b14a52d is version 1.4.2

jampoyman:
This issue also occurs when you try to view shared screens from other people in the huddle.

kciredor:
> This issue also occurs when you try to view shared screens from other people in the huddle.

Exactly, which is the reason I wanted to create an issue here, but it has already been filed ;-)

jurienhamaker:
This is turning into quite the hassle for me with my work slack. Would love to see a fix. I'll try check if I can do anything myself.

Also kind of annoying how slack pushes huddles this hard, since not everyone loves it that much, but that's a discussion for somewhere else.

jsanta:
[...]

Issue #533, opened 2022-08-02
Bug: Calls in Slack open in external browser


amura11:
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.0.0

### What Operating System are you using?

Windows

### Operating System Version

Windows 10 Proffesional 21H2 build 19044.1826

### What arch are you using?

x64

### Expected Behavior

When receiving a call in Slack the call can be accepted without having to open an external browser

### Actual Behavior

When receiving a call in Slack clicking the answer button opens the external browser

### Steps to reproduce

1. Have someone call you via Slack
2. Click the answer button

### Additional information

The `Open URLs within Ferdium` is set to true and this doesn't seem to make a difference.

0xtrsn:
I've experienced this too-- specifically for video huddles as they open in a new window and not for voice chats which stay in the same window.  The biggest problem is that, when Ferdium tries to open the new window in an external browser where you're not necessarily logged into Slack, you lose the ability to end the call as that must be done from the new window.  Makes for some quick logging in externally.

I haven't been able to test using the "Open New Windows in Ferdium" functionality, though I expect something in that line will solve the issue-- though custom, slack-specific coding may be required.

PR #2552, opened 2026-09-09
fix: keep service popups usable so Slack huddle windows open and stay visible

### LINKED ISSUE #788: Bug: Slack huddle can't be maximized
### Avoid duplicates

- [X] I have searched the [issues tracker](https://github.com/ferdium/ferdium-app/issues) for a bug report similar to mine, in vain

### Ferdium Version

6.2.0

### What Operating System are you using?

Other Linux

### Operating System Version

Fedora 37

### What arch are you using?

x64

### Last Known Working Ferdium version

_No response_

### Expected Behavior

Huddles in slack start minimized below the channels list on the left. There is a button to open the mini window. Once clicked the huddle gets an independent window that can be maximized.

### Actual Behavior

Clicking on the "open mini window" button has no effect.

### Steps to reproduce

1. Start a huddle in Slack
2. Click on the "open mini window" button that is part of the huddle (see screenshot below)

### Debug link

_No response_

### Screenshots

image

### Additional information

_No response_

### LINKED ISSUE #2299: slack requests to allow pop-ups
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that Ferdium Recipes adheres to.
- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22recipe+%3Abookmark_tabs%3A%22+label%3A%22bug+%3Abeetle%3A%22+) for a bug report that matches the one I want to file, without success.

### Ferdium Version

7.1.1

### Recipe Name

Slack

### Recipe Version

1.6.2

### Last Known working version of the recipe

NA

### Steps to reproduce

On occasion, Slack will show a pop-up at the bottom of the window that reads: "Please allow pop-ups in your browser. You can [contact Slack](http://my.slack.com/help/requests/new) if you need help."

I don't know how to allow pop-ups in a Ferdium service.

### Expected Behavior

There might be a switch in the service editor that can be activated to allow pop-ups.

### Actual Behavior

I don't know how to allow pop-ups in a Ferdium service, so nothing can be done.

### Screenshots

Image

### Additional Information

_No response_

alexcroox:
#### Pre-flight Checklist

#### Description of Change

Two independent bugs prevented Slack huddle windows from working in Ferdium. Neither is in the recipe.

**1. The popup handle never reached the page.** The service preload replaces the page's `window.open` with a shim that calls `window.ferdium.open`, a `contextBridge` function. Since #2412 the popup is created, but a `WindowProxy` cannot cross the context bridge: the page gets a detached plain-object snapshot (`[object Object]`, `document.documentElement` undefined). Slack doesn't load a URL in its huddle window; `openManagedChildWindow` opens `about:blank` with size features and renders the huddle UI into `popup.document` from the opener. That throws on the snapshot, Slack's cleanup calls `popup.close()`, and the window flashes and disappears. Slack's console shows `Attempted to window.open but received falsey return value` / `Unable to create window`.

Fix: the routing decision now happens in the page world (`src/webview/windowOpenShim.ts`). Any `window.open` call with a target name or window features uses the page's native `window.open` and returns the real `WindowProxy`; the main process handler still decides whether it becomes a popup (`new-window`) or is sent to the default browser (`foreground-tab`). A bare `window.open(url)` still goes to the default browser via `ferdium.open`, with URL objects stringified as before. The Skype-style `window.open()` placeholder keeps working, and now actually sees the page's `location.href` assignment because it lives in the page world. The preload no longer patches its own `window.open`.

**2. Joined huddles opened an invisible window.** When you join a huddle someone else started, Slack's `openHuddleWindowMva` puts Electron options in the features string: `show=no,alwaysOnTop=yes,fullscreenable=no,title=Huddle,...`. Browsers ignore these; Slack's desktop app shows the window later via its own bridge. Electron parses every `key=value` feature into a `BrowserWindow` option, so Ferdium created the huddle window hidden and always-on-top, and "Bring huddle to front" (`popup.focus()`) can't surface a window that was never shown. You hear the huddle but can't see it.

Fix: new `src/electron/popupWindowOptions.ts` normalises popup options so a service popup always behaves like a browser popup: shown, never always-on-top / frameless / transparent / kiosk / modal / skip-taskbar, fullscreenable, resizable unless the page says `resizable=no`, and a remembered off-screen position is dropped so Electron centres the window. Both window-open handlers in `src/index.ts` use it.

Other changes in `src/index.ts`:

- Popups are standalone windows rather than children of the main window, so a huddle / shared screen can be moved to another display, maximised or made fullscreen, and stays visible while Ferdium is hidden or minimised.
- Popups default to resizable. Electron 38+ made `window.open` popups resizable per the WHATWG spec; the helper from #2469 restored the old fixed-size default, which leaves Slack's huddle window unresizable because Slack never requests `resizable`.
- `foreground-tab` links inside popups open in the default browser instead of being silently dropped.

`alwaysOnTop` is deliberately not honoured: Slack asks for it because its desktop app floats the huddle window, but a web page shouldn't get an OS-level always-on-top window. "Bring huddle to front" covers that.

#### Motivation and Context

Fixes #788, #1014
Closes #2299
Related: #974, #533

Slack huddles and screen sharing have been unusable in Ferdium for years

#### Checklist

- [x] `pnpm test` passes

Tested on macOS (Apple Silicon) against a real Slack workspace: starting a huddle and opening the huddle window works, and joining a huddle started by someone else now shows the window, with "Bring huddle to front" raising it. Also verified in a minimal Electron 44 harness mirroring Ferdium's webview setup, with Ferdium's built `recipe.js` preload and Slack's exact feature strings: before, the page gets `[object Object]` and the popup is closed, and the joined-huddle window reports `visible: false, alwaysOnTop: true` even after `focus()`; after, the page gets `[object Window]`, `document.write` and rendering into the popup succeed, and the window is visible, not always-on-top, resizable and fullscreenable. Plain links, URL objects, `_blank` targets and the empty-argument placeholder all still open in the default browser. Unit tests added for the shim routing and the option sanitising.

#### Release Notes

Fixed Slack huddle and screen-share windows: popups opened by services now receive a real window handle and are always shown, so the huddle window opens and stays visible instead of flashing and closing.

### DIFF
--- a/src/electron/popupWindowOptions.ts
+++ b/src/electron/popupWindowOptions.ts
@@ -1,0 +1,88 @@
+import type { BrowserWindowConstructorOptions } from 'electron';
+
+// Chromium hands the raw window.open() features string to Electron, which
+// turns every `key=value` pair into a BrowserWindow constructor option
+// (electron/lib/browser/parse-features-string.ts). Browsers only honour a
+// handful of size and position features, but pages that also run inside
+// Electron-based desktop apps pass Electron options too. Slack, for example,
+// opens its huddle window with `show=no,alwaysOnTop=yes,fullscreenable=no,...`
+// and shows the window later through its own desktop bridge; without that
+// bridge the popup simply stays invisible (ferdium-app#788). These helpers
+// normalise the options so a popup opened by a service always behaves like a
+// popup opened in a browser.
+
+export const parseWindowFeatures = (
+  features = '',
+): Record<string, string | undefined> => {
+  const parsed: Record<string, string | undefined> = {};
+  for (const pair of features.split(',')) {
+    const [key, value] = pair.split('=').map(part => part.trim());
+    if (key) {
+      parsed[key] = value;
+    }
+  }
+  return parsed;
+};
+
+const isDisabled = (value: string | undefined): boolean =>
+  value !== undefined && /^(?:no|0|false)$/i.test(value);
+
+const parseCoordinate = (value: string | undefined): number | undefined => {
+  if (value === undefined) {
+    return undefined;
+  }
+  const parsed = Number.parseInt(value, 10);
+  return Number.isNaN(parsed) ? undefined : parsed;
+};
+
+// Popups stay resizable unless the page explicitly opts out. Electron 38+
+// made every window.open() popup resizable to match the WHATWG spec; Slack
+// opens its huddle / screen-share window without asking for `resizable`, and
+// a fixed-size window is useless for a shared screen.
+export const windowOpenFeaturesAllowResizable = (features = ''): boolean =>
+  !isDisabled(parseWindowFeatures(features).resizable);
+
+export const popupWindowOptions = (
+  features: string | undefined,
+  isPositionOnScreen: (position: { x: number; y: number }) => boolean,
+): BrowserWindowConstructorOptions => {
+  const parsed = parseWindowFeatures(features);
+
+  const options: BrowserWindowConstructorOptions = {
+    // Options a web page must not be able to control. Electron spreads these
+    // over the parsed features, so every one of them has to be set explicitly.
+    show: true,
+    alwaysOnTop: false,
+    frame: true,
+    transparent: false,
+    skipTaskbar: false,
+    focusable: true,
+    kiosk: false,
+    fullscreen: false,
+    simpleFullscreen: false,
+    modal: false,
+    opacity: 1,
+    closable: true,
+    minimizable: true,
+    maximizable: true,
+    movable: true,
+    enableLargerThanScreen: false,
+    hasShadow: true,
+    titleBarStyle: 'default',
+    // Let the user maximize / fullscreen e.g. a shared screen even when the
+    // page asked for `fullscreenable=no`.
+    fullscreenable: true,
+    resizable: !isDisabled(parsed.resizable),
+  };
+
+  // Pages remember where their popup was last time; if that position is not
+  // on a connected display any more, let Electron center the window instead.
+  const x = parseCoordinate(parsed.x ?? parsed.left);
+  const y = parseCoordinate(parsed.y ?? parsed.top);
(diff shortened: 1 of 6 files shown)


SpecialAro (changes requested):
First of all, thank you so much for this PR. I was working on something like this already. Glad I didn't waste more time on it :)

I don't think we should honor `resizable=no` in `popupWindowOptions()`.

Electron 44.2.0 explicitly discards the `resizable` feature (https://github.com/electron/electron/blob/aa650d74597c652878629df0038a50485e156a09/lib/browser/parse-features-string.ts#L163-L167) when parsing `window.open()` feature strings. In Electron's `parseFeatures()`, it states that per spec windows are always resizable and then removes `parsed.resizable` before constructing the parsed `BrowserWindow` options.

This is also the parser Electron uses when creating guest windows: `openGuestWindow()` passes the page-provided `features` string through `parseFeatures()` and spreads the resulting `parsedOptions` into the `BrowserWindow` options.

So with Electron 44.2.0:

```text
window.open(url, target, 'resizable=no')
```

does **not** result in `BrowserWindow({ resizable: false })`.

By handling it ourselves with:

```ts
resizable: !isDisabled(parsed.resizable),
```

Ferdium would diverge from Electron/browser semantics and reintroduce support for a feature Electron deliberately ignores. A service carrying a legacy `resizable=no` value could therefore get a fixed-size popup in Ferdium when the equivalent native Electron popup would remain resizable.

Could we remove the `resizable` Electron Webview option entirely and remove/update the `resizable=no` test accordingly?

SpecialAro:
Took the liberty of making the changes myself :)

SpecialAro (approved):

SpecialAro:
@all-contributors please add @alexcroox for code

alexcroox:
thanks for your help!

